PR #1, opened 2014-08-26
CommandBar by ID

bpatra:
There are two CommandBar with name "Cell": one for the usual right click on a cell and one for the "PageBreakView" [see this article](http://msdn.microsoft.com/en-us/library/office/gg469862(v=office.14).aspx). When passing to the method *ExcelCommandBarUtil.LoadCommandBar* an xml argument of the form

    <commandBars xmlns='http://schemas.excel-dna.net/office/2003/01/commandbars' > 
      <commandBar name='Cell'>
      <button caption='Hello' enabled='true' onAction='SayHello'/>
      </commandBar>
    </commandBars>"

the first CommandBar whose name is "Cell" is returned. We have found situations where this first CommandBar is the "Cell" for PageBreakView and not the usual one.

In this pull request we add the possibility to specify the id of the CommandBar to avoid such situations. If an attribute *id* is present then it will be used to retrieve the CommandBar instead of *name*. If no *id* attribute is present then the *name* is used as a fallback (to avoid breaking existing code). Note that the attribute *name* is mandatory in both cases.

Now if you want to retrieve the usual "Cell" CommandBar you can use its *id* 424.

    <commandBars xmlns='http://schemas.excel-dna.net/office/2003/01/commandbars' > 
      <commandBar name='Cell' id='424'>
      <button caption='Hello' enabled='true' onAction='SayHello'/>
      </commandBar>
    </commandBars>"

### DIFF
--- a/Source/ExcelDna.Integration/ExcelCommandBars.cs
+++ b/Source/ExcelDna.Integration/ExcelCommandBars.cs
@@ -38,29 +38,6 @@ namespace ExcelDna.Integration.CustomUI
 {
     public delegate Bitmap GetImageDelegate(string imageName);
 
-    //public class ExcelCommandBars
-    //{
-    //    public ExcelCommandBars()
-    //    {
-    //    }
-
-    //    public virtual string GetCustomUI()
-    //    {
-    //    }
-
-    //    public virtual object GetImage(string imageName)
-    //    {
-
-    //    }
-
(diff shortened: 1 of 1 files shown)


govert:
OK - this is finally integrated. Sorry it took so long - I'm moving the main repository over to GitHub at last.

Issue #17, opened 2015-08-28
Replace the post-build commands with custom MSBuild tasks + custom build target


caioproiete:
As discussed in #15 we should write one or more custom MSBuild tasks, ship them into one assembly, and have a custom build target to run them as part of the build.

These MSBuild tasks would be responsible for copying the XLL to the bin folder, as well as running ExcelDnaPack based on a conditional compilation symbol.

govert:
I think this should be done with custom MSBuild targets, but using the ExcelDnaPack binary rather than a separate library of build tasks.

Is there any reason not to implement the build targets that way? Keeping the ExcelDnaPack as an executable that does the extra work means that other approaches to build automation are still supported.

CaioProiete:
I think we should have both. A custom build target that uses an MSBuild task to perform the job (which would be included in the NuGet package), and a separate utility (ExcelDnaPack) distributed outside of the NuGet package, for people who want to run it via command-line.

Both the ExcelDnaPack utility and the MSBuild task assembly would share the exact same code, but with just different entry points.

I think from a build perspective, it is better to have a tighter integration with MSBuild vs. spinning up a new process every time you build (which is harder to guarantee that it will actually stop, and it is much harder to raise proper error messages and fail the build when a validation fails or something goes wrong - e.g. unable to patch the version because the attribute is missing in the .dna file).

ps: I'm sure I'm being a little pedantic - sorry about that :)

caioproiete:
This has been fixed with merging of #27

Issue #15, opened 2015-08-27
Automatically replace XLL version information with data read from an assembly


caioproiete:
This is a (slightly modified) feature request that was asked in the mailing-list, so just adding as an issue so we can track this better and possibly encourage someone to implement and send a PR.

Original requests:
- [Exposing resource info in ExcelDNAPack](https://groups.google.com/forum/#!topic/exceldna/GusDyS0Sf0U)
- [Versioning the xll for Windows Installer](https://groups.google.com/forum/#!topic/exceldna/HlVeZUZNcqM)

My assumption is that the developer (or ideally the continuous integration system) should be responsible for setting the right version information about the add-in, in the main assembly (or in all of them). With that said, we should be able to describe in the Excel-DNA XML file which assembly should be used to read version information about the add-in (version number, copyright, product name, etc.) and any information available in this assembly should automatically replace what is in the XLL file, during the build of the project.

I would suggest this to be implemented as a regular MSBuild Task, rather than implementing this as part of the ExcelDnaPack utility given that people might not care about packing their add-ins - and still would like to get version information in the XLL that gets copied during the build.

govert:
I think it's a good idea, but not something I'm likely to work on myself anytime soon.

Is there a standard MSBuild task that can do a version update for a .dll?

If not, I think the code for the version update would fit well into the ExcelDnaPack executable, since that already modifies the .xll file. Maybe ExcelDnaPack should have an option to update the Version resource only (without packing the .dna file). That would eliminate the need for another binary to update the Version resource, but still give you the option to only do a version update.
I'm happy that a build step might call ExcelDnaPack for the version update, or even a custom built target (though I wouldn't know how to make such a thing).

I would like to discuss details of the .dna changes before you implement that aspect. What attributes do you plan to add etc...

CaioProiete:
Hi Govert,

I understand this is not in your short-term roadmap, which is why I added the issue here to surface it to others as you said you'd be open to a PR (I'll take a stab on this in a few days, if no one does).

I'm not aware of any built-in MSBuild task that can do that. I was thinking about implementing a custom one that would be included in the NuGet package and run as part of the build (via a custom build target). This MSBuild task would just modify the version resource of the .XLL file(s) to set the corresponding version info, all using Windows APIs, similar to what you already do in ExcelDnaPack.

[...]

PR #26, opened 2015-08-30
Fix ExcelDnaPack crashing due to unhandled exceptions | fixes #25

### LINKED ISSUE #25: ExcelDnaPack crashes during build - needs to catch exceptions before MSBuild does
When an unhandled exception occurs in `ExcelDnaPack`, it lets it bubble up to the caller, and crashes. This causes MSBuild/Visual Studio to show the dialogs below:

pack-error-01
pack-error-02

We need to catch any unhandled exceptions in Main, and return a Environment.ExitCode >= 1 to signal that it failed - so that MSBuild can fail the build too.

CaioProiete:


govert:
The last time I fixed up the exit codes, if was just a quick fix and never properly cleaned up. Instead of just adding the exception handler, I'd prefer if we can at least clean up a bit too. I suggest:

* Your new Pack method should rather return some result code (maybe a new enum) from all the `return;` exit locations, and not set `Environment.ExitCode` itself. 

* Then the new Main becomes the only place where the process exit code is set. There are some places which call `Environment.Exit(1)`. These should probably throw an exception, which will be caught and displayed in Main.

* We must decide on the result codes returned from the process. At the moment it seems like a mix of -1, 0 and 1. I don't have a strong opinion on whether all error just return 1 (or -1), or whether we return different error codes. But the current mix of sometime returning -1 and sometimes 1 for the error seems weird. I'd be OK with all error conditions returning 1, if you think that makes sense.

CaioProiete:
@govert: Absolutely agree with all the points above, but I'm currently working on #17 (made great progress today BTW, expect to have a PR tomorrow), and `ExcelDnaPack` crashing is a blocker as I cannot reliably fail the build when something goes wrong.

This PR does the bare minimum change to fix the issue, without introducing any breaking changes - so it actually makes things better than it is today / moves us forward.

I'd prefer you merge this one as is, and open a new issue for improving `ExcelDnaPack` exception handling, and we can revisit this later (likely as part of #15).

govert:
OK - then let's put all of it into a temporary branch, at least until I'm done with v0.33. Maybe "build-targets" or something.
Do I need to make the branch, or can you do that as part of the Pull Requests?

CaioProiete:
Great.

I don't have write permission on the main repo, so you'll have to create the branch. I can only open pull-requests for existing branches.

govert:
Done.

Issue #18, opened 2015-08-28
Debug instructions to run Excel should be stored in the project file


caioproiete:
When I install the Excel-DNA NuGet package, it adds the instructions to allow me to debug the add-in by starting Excel and passing the add-in file as a parameter, which is great.

![debug](https://cloud.githubusercontent.com/assets/177608/9558472/7360c46c-4dbc-11e5-91d6-fa829d5ff9c0.png)

Unfortunately it doesn't make this change in the project file (e.g. `.csproj`). Instead, it updates the user's file related to the project (e.g. `.csproj.user`) which is individual for each developer, and therefore is not committed to the source control.

Example of a `.csproj.user` file:

```
<?xml version="1.0" encoding="utf-8"?>
<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/ ...">
  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|AnyCPU'">
    <StartAction>Program</StartAction>
    <StartProgram>C:\Program Files\Microsoft Office\Office15\EXCEL.EXE</StartProgram>
    <StartArguments>"MyExcelAddIn-AddIn.xll"</StartArguments>
  </PropertyGroup>
</Project>
```

In practice, what happens is that **when a developer clones a repository from source control and tries to debug it, it never works**...

The workaround so far, is to manually copy/paste the contents of the `.csproj.user` file into the `.csproj` file.

Example of a manually changed `.csproj` file:

```
<?xml version="1.0" encoding="utf-8"?>
<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <!-- ... -->
  <PropertyGroup>
    <!-- ... -->
    <TargetFrameworkVersion>v4.5.1</TargetFrameworkVersion>
    <FileAlignment>512</FileAlignment>
    <StartAction>Program</StartAction>
    <StartProgram>C:\Program Files\Microsoft Office\Office15\EXCEL.EXE</StartProgram>
    <StartArguments>"MyExcelAddIn-AddIn.xll"</StartArguments>
  </PropertyGroup>
  <!-- ... -->
```

---

What I would like, is for the NuGet package to update the `.csproj` file with the `StartAction`, `StartProgram` and `StartArguments`, instead of updating the `.csproj.user` file (and equivalent for other languages).

govert:
[...]

Issue #19, opened 2015-08-29
Improve the way we determine the path for Excel and add-in platform, for debugging


caioproiete:
This is somewhat related to issue #18, where I'm proposing that we move the Excel.exe path and add-in to the project file, so that the debugging information is persisted in source control, and any developer that clones the repository would then be able to debug the add-in without having to configure anything.

We know that we can't control what version of Excel each developer has installed on his/her machine... Some developers run Excel 2010 32-bit, others run Excel 2010 64-bit, others Excel 2013 64-bit, etc.

I'd like a developer to be able to clone my repository, open my solution in Visual Studio, and just hit F5 and be able to Debug it, starting an instance of the Excel installed on his/her machine, no matter what version is there, and with the right add-in XLL file (x64 or x86) for that debugging session.

---

Whilst I don't have a perfect solution for that yet, I do have a workaround that have been working well for most cases:
Inside the project file, I add a few conditional property groups that define the `StartProgram` variable with the correct path of the Excel on the machine (if matches any, of course)

```
<PropertyGroup Condition="'$(StartProgram)' == '' AND Exists('C:\Program Files\Microsoft Office\Office15\EXCEL.EXE')">
  <StartProgram>C:\Program Files\Microsoft Office\Office15\EXCEL.EXE</StartProgram>
</PropertyGroup>
<PropertyGroup Condition="'$(StartProgram)' == '' AND Exists('C:\Program Files\Microsoft Office\Office14\EXCEL.EXE')">
  <StartProgram>C:\Program Files\Microsoft Office\Office14\EXCEL.EXE</StartProgram>
</PropertyGroup>
<!-- etc... -->
```

I also have the similar conditionals for 32-bit Excel installed on Windows 64-bit machines:

```
<PropertyGroup Condition="'$(StartProgram)' == '' AND Exists('C:\Program Files (x86)\Microsoft Office\Office15\EXCEL.EXE')">
  <StartProgram>C:\Program Files (x86)\Microsoft Office\Office15\EXCEL.EXE</StartProgram>
</PropertyGroup>
<PropertyGroup Condition="'$(StartProgram)' == '' AND Exists('C:\Program Files (x86)\Microsoft Office\Office14\EXCEL.EXE')">
  <StartProgram>C:\Program Files (x86)\Microsoft Office\Office14\EXCEL.EXE</StartProgram>
</PropertyGroup>
<!-- etc... -->
```

[...]

Issue #16, opened 2015-08-28
Build event commands should not use hard-coded paths


CaioProiete:
They should use macros to determine the path based on the project (or solution path).

![build-event](https://cloud.githubusercontent.com/assets/177608/9545156/4f7092da-4d5d-11e5-8fe7-2915580c835c.png)

This would enhance the experience of developing an Excel add-in with a team or developing an open-source add-in, as any person would be able to clone the repository and compile the project, no matter what folder structure they used.

govert:
I think this is already fixed. Can you check again with a new project and adding the pre-release NuGet package:
    
    > Install-Package -Pre ExcelDna.AddIn

There is one place where the path is hard-coded, and that is in the Debug command line. I am not sure how to set the working directory to be the Output folder, and I thought Excel would sometimes use the wrong directory. So I added the full path there. But now I'm not sure whether that is needed, or how to find out. One user had ".\" as the Working Directory, but is that different to the default?

CaioProiete:
Thanks @govert.

I've just tested with `v0.33.7-rc1` from the NuGet feed and it is not fixed there, at least.
![build-event-2](https://cloud.githubusercontent.com/assets/177608/9558105/3c23dc90-4db8-11e5-98ff-22e5c29d87d2.png)

Regarding the Debug, what I've found is that when you ask Visual Studio to debug an application, it automatically sets the working directory to `bin\Debug` (or `bin\Release` depending on the build config selected), so if you just put the name of the XLL file without any path, it just works.

Here is a screenshot of how I configure all my add-ins:

![start-program](https://cloud.githubusercontent.com/assets/177608/9558167/cd612410-4db8-11e5-94d1-3408720ea770.png)

govert:
OK, so I create a new Class Library project, and then do 
`PM> Install-Package -Pre ExcelDna.Addin`

Then I get:
![image](https://cloud.githubusercontent.com/assets/414659/9558265/dbb748a2-4de3-11e5-8198-dc7f6dd525e4.png)

What are you doing?

interdrift:
This is indeed something that would be very useful. I had a colleague of mine download my code and it turned out he couldn't compile it until he uninstalls excel-dna package and download it again from nuget.

govert:
@CaioProiete I've looked at the code for the Install.ps1 here: https://github.com/Excel-DNA/ExcelDna/blob/master/Package/ExcelDna.AddIn/tools/install.ps1

[...]

PR #27, opened 2015-09-03
Replace post-build events with MSBuild targets + improvements

caioproiete:
Hi Govert, so here it goes the PR as promised.

On this PR I tried to focus on fixing #17 and whilst I was at it, I've added a few improvements that I thought would make sense, and ended up fixing #16 as well. This PR also opens the door for #15, #18 and #19.

### What does it do / How it works
This PR adds two new files to the NuGet package:

#### tools\ExcelDna.targets
This is a vanilla MSBuild project file that contains all the logic to replace what currently is done via post-build events. A link to import this file gets added to the `.csproj` / `.vbproj` / `.fsproj`, etc. so it gets loaded together with the project.

MSBuild files are quite verbose in general, but the code is pretty straightforward and does what used to be in the post-build events (and more things which I explain later in this document):

  * Copy the -AddIn.dna file to the output folder
  * Copy the 32-bit .xll, with the same name as the -AddIn.dna, to the output folder
  * Run ExcelDnaPack for the -AddIn.dna file
  * Copy the -AddIn.dna file to a new file -AddIn64.dna to the output folder
  * Copy the 64-bit .xll, with the same name as the -AddIn64.dna, to the output folder
  * Run ExcelDnaPack for the -AddIn64.dna file

Among other things, this MSBuild file has a target called "ExcelDnaBuild" which hooks in to the AfterBuild event and therefore is is triggered every time a project is built, keeping a similar behavior as it is in v0.32.

The developer is not expected to edit or interface with the `ExcelDna.targets` file. It lives inside the nuget package folder, and gets replaced every time the NuGet package gets updated.

#### content\ExcelDna.props
This is another vanilla MSBuild project that gets added to the VS project and allows the developer to customize the default behavior of the ExcelDna build process.

This seemed a good way to give back some control/flexibility to the developer, now that the post-events are gone. So it is possible to configure things like:

* Enable / Disable processing of platform-specific .dna files (e.g. don't build 32-bit files)
* Enable / Disable generation of platform-specific .dna files (e.g. don't generate 64-bit files)
* Enable / Disable packing of .dna files

So in other words, the `ExcelDna.targets` define a number of defaults, which match the current behavior of v0.32, and the `ExcelDna.props` allows the developer to override pretty much every aspect of it, if he/she wants to. The file that gets added is just a template with the variables that I think will be more commonly changed, but a more advanced developer can extend it to override almost every variable in the `.targets` file and also has the opportunity to hook into the build process to execute custom steps.

So, in summary, `ExcelDna.props` is there for the developer to edit it as he/she see fit. But if it gets renamed or deleted from the project, it's perfectly fine and the default values defined in the `ExcenDna.targets` will be used (which, again, match current v0.32 behavior).

### Improvements / New Features
Being able to hook into the MSBuild pipeline gave me the chance to make some improvements that I find useful, and maybe others will do too.

#### Clean Solution
I've introduced a target that hooks into the "Clean Solution" target (which also fires on Rebuild by default) which will delete any .dna, .xll, or .xll.config files in the output folder (e.g. `bin\Debug`).

#### Build Configuration File
As mentioned above, this adds a file called `ExcelDna.props` to the project that allows the developer to turn on/off certain steps of the build process. These parameters can also be overridden via the MSBuild command-line.

This is useful, for example, if during development you only care about building 64-bit versions of the add-in without packing, but then in the build server you want to build both 32-bit and 64-bit versions and pack them all.

#### Support for multiple .dna files in the same project
The build target discovers any .dna file that exists in the project (including sub-folders) and automatically copies them to the output folder.

It also detects (based on conventions) what is the platform of the .dna file (e.g. 32-bit - `MyAddIn.dna`) and generates the corresponding version (e.g. 64-bit - `MyAddIn64.dna`). It also works the other way around (e.g. if you only have a file called `MyAddIn64.dna`, it will generate a `MyAddIn32.dna`). All of this is configurable via `ExcelDna.props`, so you can disable the automatic generation of any of these.

The build target also honors any folder structure you may have in the project, so if you have .dna files in sub-folders, they will be copied to the output directory using the same structure, as you'd expect.

#### Support for custom platform-specific files
There are cases where the the 32-bit .dna file and the 64-bit .dna file need to reference different assemblies, therefore automatic generation is not an option. So if the project has a single file called `MyAddIn.dna`, it understands that this is a 32-bit file and (by default) will generate a `MyAddIn64.dna` file by making a copy of the 32-bit version.

However, if the project has two files `MyOtherAddIn.dna` and `MyOtherAddIn64.dna`, the build target will respect that and will not overwrite either of these files with a copy from another one. It will use the files exactly as they are in the project.

#### Support for any naming convention
Because all .dna files as discovered dynamically, the files don't have to be `ProjectName-AddIn.dna` and a developer call it anything (e.g. `MyAwesomeXLApp.dna`). The only important part is the suffix of the file, which determines the platform (e.g. `MyAwesomeXLApp.dna` = 32-bit, `MyAwesomeXLApp64.dna` = 64-bit).

The suffix is also configurable via ExcelDna.props, so a developer can use whatever suffix he/she wants (e.g. `MyAwesomeXLApp-x86.dna` = 32-bit, `MyAwesomeXLApp-x64.dna` = 64-bit).

#### Support for App.config files (specific or generic) - if any exist
The build target will try to discover the most appropriate App.config file in the project and will copy it to the output folder, renaming it to match the corresponding .xll file.

Some examples:
* If the only config file in the project is called `App.config`, then any add-in will get a copy of it
```
App.config => MyAddIn.xll.config
App.config => MyAddIn64.xll.config
```

* If there are specific files `MyAddIn.config` and `MyAddIn64.config`, then it will use that instead (even if an App.config exists)
```
MyAddIn.config => MyAddIn.xll.config
MyAddIn64.config => MyAddIn64.xll.config
``` 

* Also, if you have .dna files in sub-folders, it will give priority to the `App.config` that is in the same folder as the .dna file, and then fallback to the root folder of the project to use the `App.config` there, if exists.

### Breaking Changes
Hopefully none.

I tried very hard to keep the existing behavior and default settings, and my tests show that upgrading add-ins from v0.32 to this new version works exactly the same way for C#, VB .NET, and F# projects (I've tested with VS2010 from .NET 2.0 to .NET 4.0 and VS2013, from .NET 2.0 to .NET 4.5.1).

#### Install.ps1

* (Obviously) Removed the code that sets the post-build event commands

* Removed the code that sets the `CopyToOutputDirectory` and  `BuildAction`, because it is not needed anymore. The MSBuild target now discovers .dna files in the project that have `BuildAction` `None` or `Content`

#### Uninstall.ps1

* (Obviously) Removed the code that cleaned the post-build event commands (no more desserts, banana, etc. sorry :))

* Also removed the code that sets the `CopyToOutputDirectory` and `BuildAction` for the same reason above in Install.ps1

### Next steps
Hopefully this is in the right direction and you will take it for a spin - and hopefully not find major bugs.

I'd like to get this PR merged, and then start working on #15 which I believe will allow me to simplify the `.targets` a lot - you see, in this PR I'm using only native MSBuild tasks which makes things more difficult for simple things like loops and merge of collections as you can see in the code. My idea with #15 would be to create a custom MSBuild task for doing the version change and packing, and I can take the opportunity to create some other useful tasks that will simplify the MSBuild code.

Looking forward to get your feedback.

Thanks,
Caio Proiete

ps: This PR works better if you merge #25, but it is not required for you to test and validate that it is doing what you'd expect.


govert:
That looks awesome! I'll have a closer look and try to figure out how to deal with the branching.

govert:
@CaioProiete About the .props file:
* Do you have some reference on how these are supported? I only run into stuff related to C++ projects.
* Could the .props file be called .targets (different to the .targets in the package directory)?

govert:
OK - no skip those questions about the .props file - I think it's fine.

govert:
I would like to compare your approach using the Install.ps1 to add the .targets file, to the NuGet 2.5 approach: https://docs.nuget.org/release-notes/nuget-2.5

CaioProiete:
@govert: The `.props` file is just a normal `.targets` file and can have any extension we want. The `.props` extension is just a convention that Microsoft uses for files that only define properties (and not targets).

The NuGet 2.5 approach does exactly the same thing I'm doing via Powershell. I didn't go with the NuGet 2.5 approach because I thought you needed to support old versions of NuGet (given you support .NET 2.0).

Package/ExcelDna.AddIn/tools/ExcelDna.targets
```diff
@@ -0,0 +1,291 @@
...
+    <!--
+      Make sure the ExcelDnaPack utility exists
+    -->
+    <Error Text="ExcelDnaPack.exe not found: $(ExcelDnaPackExePath)"
+           Condition="!Exists('$(ExcelDnaPackExePath)')" />
+
+    <!--
+      Make sure the suffix for 32 and 64-bit add-ins are different, so we don't get clashes
+    -->
+    <Error Text="ExcelDna32FileSuffix and ExcelDna32FileSuffix cannot be identical. Fix you ExcelDna.props file"
```
govert: Error text should be ExcelDna32FileSuffix and **ExcelDna64FileSuffix** cannot be identical. Fix **your** ExcelDna.props file

govert:
@CaioProiete OK - I've been able to install it in a new C# project, and every works.

It's fantastic - thank you!

I have a few things to think about:
* Should the file be called "ExcelDna.AddIn.targets", to match the package name? This is the NuGet 2.5 convention.
* Should the file be called MyLibrary.props or MyLibrary-AddIn.props, to make clear that the .props file is part of the project? I'm not sure it matters, I'm just wondering what I'd like to see in the Project file list.
* I'm confused by the ExcelDnaBuild64Dna vs ExcelDnaCreate64DnaFrom32Dna flags. The former should perhaps be called ExcelDnaBuild64AddIn. I don't know what to think about the ExcelDnaCreate64DnaFrom32Dna and ExcelDnaCreate32DnaFrom64Dna flags. My first attempt to not build the 64-bit add-in failed, because I just set ExcelDnaBuild64Dna to false, but that was not enough.
* Maybe we want to configure more than just the packed add-in's suffix?
* I'm confused by the .dna discovery and naming story, but have not looked too closely at that yet. Do you use renamed or nested .dna files?

Going forward:
* I've checked in the changes I wanted for version 0.33, including the install.ps1 changes. Maybe you can give that a try too?
* Would it make sense to make a v0.33 branch for any other changes before the release, and then merge this into master as the start of the next version?
* I think you should not go too fast in moving the build tasks into (or out of) ExcelDnaPack, or refactoring ExcelDnaPack. For now I'm actually quite happy with the stuff in the .targets file. Adding a version update feature to ExcelDnaPack as discussed in #15 would be great. I'm sure after a while I'll be happy with refactored build tasks and tools.

It's a great hob, thanks again for the effort so far!
I'll look more closely and post questions and comments as I go along.

Package/ExcelDna.AddIn/tools/ExcelDna.targets
CaioProiete: Thanks. Will fix it.

CaioProiete:
Thanks @govert! I'm glad you like it.

> Should the file be called "ExcelDna.AddIn.targets", to match the package name? This is the NuGet 2.5 convention.

Sure. If that is the convention, let's follow it. I'll update this PR.

> Should the file be called MyLibrary.props or MyLibrary-AddIn.props, to make clear that the .props file is part of the project?

I'm not sure about the "MyLibrary" part. Seems too generic. This file is specific to the configuration of the ExcelDna build process, so I think the name should be more explicit. So following the convention above, maybe "ExcelDna.AddIn.props" would be more appropriate. Also the developer might install other NuGet packages that could bring other `.props` files and we def. don't want the risk of a name clash.


> I'm confused by the ExcelDnaBuild64Dna vs ExcelDnaCreate64DnaFrom32Dna flags.

When you set `ExcelDnaBuild64Dna` to `false` you are basically saying: Ignore any 64-bit .dna files present in the project and pretend they don't exist (so any *64.dna files will be skipped when you build the solution/project).

Similarly, setting `ExcelDnaBuild32Dna` to `false` means ignore .dna files that are detected are being 32-bit, and do nothing with them.

So in your test, I suspect you installed the NuGet package and had a single "MyProject.dna". By convention, this is detected as being a 32-bit file, so the property `ExcelDnaBuild64Dna` has no effect on it...

If you rename this file to "MyProject64.dna", though... Then it will be skipped during the build, because you disabled building 64-bit files.

---

On the other hand, we have something else that happens as part of the build process, which is the automatic generation of .dna files for the alternative platform (contrary to the .dna being built).

So if I see a file called `Govert.dna`in the project, I detect it as being 32-bit (again, because of naming convention) and I copy it to `Govert64.dna` along with corresponding .xll, etc. If you don't want me to generate this this 64-bit file automatically for you, then you need to set the property `ExcelDnaCreate64DnaFrom32Dna` to false.

Similarly, if I see a file called `Govert64.dna`in the project, I detect it as being 64-bit (once again, because of naming convention) and I copy it to `Govert32.dna` along with corresponding .xll, etc. If you don't want me to generate this this 32-bit file automatically for you, then you need to set the property `ExcelDnaCreate32DnaFrom64Dna` to false.

---

I thought the names of the properties (along with the comments) were clear enough:
* ExcelDnaBuild32Dna => ExcelDna Build 32 Dna
* ExcelDnaBuild64Dna => ExcelDna Build 64 Dna
* ExcelDnaCreate64DnaFrom32Dna => ExcelDna Create 64 Dna From 32 Dna
* ExcelDnaCreate32DnaFrom64Dna => ExcelDna Create 32 Dna From 64 Dna

But based on your comments I'm not so sure. Any suggestions for better naming are welcome.

**Important**: Remember that one thing is having the flexibility to configure several aspects of the build process - and I allow this by giving the developer the opportunity to override a number of properties that are referenced in the `.targets`.

The `.props` file, however, is meant to be a starting point of configuration, containing the properties that the developer is more likely to change. So if you think that `ExcelDnaBuild64Dna` vs `ExcelDnaCreate64DnaFrom32Dna` causes more confusion than it helps, then I suggest we just remove one of them from the `.props` file (`ExcelDnaBuild64Dna` is probably a better candidate for removal) but still leave the possibility for the developer to add it to the `.props` file later if he/she wants to control that aspect of the build.

And this is true for all the properties that have default values in the `.targets` file. A (likely more advanced) developer could open the `.targets` see what he/she wants to override, and add to the `.props file`.


> The former should perhaps be called ExcelDnaBuild64AddIn.

I'm indifferent here... Happy to rename it if you think it will make it easier. Let me know.


> Maybe we want to configure more than just the packed add-in's suffix?

Not sure about that... The .dna file probably is what you want your add-in to be named in the end, and this PR gives you the ability to use any name you want, so I think 99.9% of the devs will only care about the suffix. I say we leave as is, and revisit this if we get any requests.

It is pretty easy to work around that, though... All the developer would have to do, is add a new target to be executed after the `ExcelDnaBuild` target, and rename the file in the output folder.


> I'm confused by the .dna discovery and naming story, but have not looked too closely at that yet. Do you use renamed or nested .dna files?

I have scenarios where one .csproj contains more than one .dna files and sometimes inside sub-folders.
I also have scenarios where my 64.dna file cannot be the same as the 32.dna file, as they reference different DLLs that are platform-dependent, and they also have their separate .config files.

But in the end, what I think is the most useful, is to have freedom to call the .dna files whatever you want, independently of $(ProjectName), put it anywhere in the project (root or sub-folder) and it "just works". This (I think) a lot of people will benefit from. Having multiple .dna files in the same project, probably not so frequent.


> I've checked in the changes I wanted for version 0.33, including the install.ps1 changes. Maybe you can give that a try too?

Sure, will do. Will also rebase on top of that change to make sure it integrates well with it.

> Would it make sense to make a v0.33 branch for any other changes before the release, and then merge this into master as the start of the next version?

Yes, that would be great, so we can get this PR (and others) merged into master sooner, and you still have the chance to patch bugs that are critical for v0.33.


> I think you should not go too fast in moving the build tasks into (or out of) ExcelDnaPack, or refactoring ExcelDnaPack. For now I'm actually quite happy with the stuff in the .targets file. Adding a version update feature to ExcelDnaPack as discussed in #15 would be great. I'm sure after a while I'll be happy with refactored build tasks and tools.

I understand you are concerned with introducing big changes and #15 could be one of them, but (after some spikes) I don't see how we could resolve #18 and #19 without a custom MSBuild task, so not being able to move forward with #15 using a custom task is a blocker for #18 and #19.

Would you be OK with adding another `.dll` in the NuGet package, to contain the MSBuild tasks required for #18 and #19? If you are, then I can leave `ExcelDnaPack` alone for a while until you think you're ready, and I could even add another MSBuild task for #15 in the same `.dll`.

govert:
Let's move the future tasks discussion to #15.

---

With "MyLibrary.props or MyLibrary-AddIn.props" I meant $(ProjectName).props or $(ProjectName)-AddIn.props, like the .dna file that is generated by default.

My reasoning is like this:
* When adding a .props and .targets to the NuGet package (in the NuGet 2.5 style), these will stay in the NuGet package and not be modified by the user.
* Our use of ExcelDna.targets is exactlyt like this (though I'd like it renamed to ExcelDna.AddIn.targets for consistency).
* Our use of ExcelDna.props is different - it's added to the project, with the intention that the user check it into source control, that it appear in the project tree and that it be modified by the user.
* So in many ways it's like the .dna file - even though applied to the building. So maybe $(ProjectName)-AddIn.props?
* Another way of seeing it (and the .dna file) is like App.config or AssemblyInfo.cs. That might suggest a project-independent name.
* What about putting the .props file under the Properties folder, next to AssemblyInfo.cs. In a way it's meant for advanced use anyway.
* I don't feel so strongly about this, I just want to take the opportunity to think about this a bit while we're busy - and I have your attention;-) I understand we can change how this works later, at least until a release that uses it.

--- 

I think my confusion with "ExcelDnaBuild32Dna" is that I don't see the .dna file as the thing being 'built'. I think we're building an add-in, (maybe two). So I'm pretty sure this should be called something else, like "ExcelDnaBuild32AddIn" and "ExcelDnaBuild64AddIn", or "ExcelDnaBuild32BitAddIn" and "ExcelDnaBuild64BitAddIn", and that those flags should do exactly what they say - switch on and off the building of the two add-ins.

I don't want ExcelDnaCreate64DnaFrom32Dna and ExcelDnaCreate32DnaFrom64Dna in the default .props file at all. It's too confusing. It can be a special flags that advanced users can add. maybe. 

Maybe the way these flags interact with ExcelDnaBuild64AddIn in the .targets file should be sorted out a bit too. Certainly from the default install, a single flag change in the .props file should disable the 64-bit build completely.

--- 

OK, I'm happy with the flexible .dna file discovery story if you think it's useful.

-- 

I guess we also need to say something in the NuGet package's readme.txt file.

CaioProiete:
Thanks @govert. I've just updated the PR with some of the changes discussed above (details below).

#### Renaming the .targets file to "ExcelDna.AddIn.targets"
This is done. I've updated the PR.


#### MyLibrary.props, etc.
Got it. It was a late reply and I guess I was asleep :) and didn't catch you meant `$(ProjectName)`.

I think `$(ProjectName).props` is a little dangerous because we don't know if the developer already has a file like that in his/her project (because of his own custom properties, or another NuGet package that added it), so I'd be more in favor of `#(ProjectName)-AddIn.props` to minimize the chance of a name clash (or any other kind of convention that appends a suffix).

At the same time, I think some people could get confused with `$(ProjectName)-AddIn.props` for cases where they have two .dna files in the project (e.g. `MyLibrary.dna` and `MyLibrary64.dna`).

I can see some questions a dev could have in these scenarios:
* Does `MyLibrary-AddIn.props` gets used for both 32 and 64-bit?
* Can I add `MyLibrary-AddIn64.props` for properties that affect only 64-bit add-ins?

So I would prefer a more project-independent name, such as `ExcelDna.AddIn.props`, to keep the consistency with the `.targets`.

I like the idea of moving the `.props` file to the `Properties` folder. I've updated the PR with that, see what you think.

Whatever naming we agree, I'd recommend we stick to a static name, rather than a dynamic name, as if we use a dynamic name, we can't rely on NuGet do do the work of adding/removing on install/uninstall, and will have to handle this in the Powershell scripts, similar to what you currently have to do for the `ExcelDna-Template.dna`.


#### ExcelDnaBuild32Dna and/or ExcelDnaCreate64DnaFrom32Dna
Leaving naming of variables aside, I think we need to get an understanding of the behavior we want after a fresh install of ExcelDna, and what will be possible to be tweaked via .props file (even if the property is not there by default).

Imagine scenario where your .csproj looks like this (this is real scenario BTW):

* ModuleA.dna
* ModuleB.dna
* ModuleB64.dna
* ModuleC64.dna

By default, this is what will happen:

* All 4 .dna files will be copied to `bin\Debug`, together with their corresponding xll files (discovered based on platform).

This process (to me) is the "build" process (or call it whatever you want, but this is "step 1"). If they were .cs files, I would get a .dll in `bin\Debug`. Because they are .dna, I get an .xll instead, along with the .dna, of course.

* All 32-bit .dna files that are "alone" and don't have a 64-bit "pair" will get one automagically. And this 64-bit generated file will be a copy of the 32-bit one. Likewise, all 64-bit .dna files that are "alone" will get a 32-bit "pair" generated.

That means we will copy `ModuleA.dna` to `bin\Debug\ModuleA64.dna`, and copy `ModuleC.dna` to `bin\Debug\ModuleC64.dna`, together with their corresponding xll files (discovered based on platform).

This process (to me) is the "generate" process  (or call it whatever you want, but this is "step 2")

---

So as you can see, these are effectively two different things happening, and I think a developer should have the ability to enable/disable any of these two steps.

`ExcelDnaBuild32Dna` & `ExcelDnaBuild64Dna` gives you control over step 1.

`ExcelDnaCreate64DnaFrom32Dna` & `ExcelDnaCreate64DnaFrom32Dna` gives you control over step 2.

Now, I'm sure we can come up with better names for these variables (and we'll have to document this, of course), and it probably makes sense that we leave only one of these pairs of variables in the .props file to avoid confusion.

But I think we need this type of granular control.

So what I propose (which I think would address the confusion) is this:
The .props file by default only includes `ExcelDnaCreate64DnaFrom32Dna` and `ExcelDnaCreate64DnaFrom32Dna` - which I have renamed to `ExcelDnaAutoGenerate64BitAddIn` and `ExcelDnaAutoGenerate32BitAddIn` in an attempt to make it more clear.

See what you think.

govert:
OK - what do you think about the fixed name "ExcelDna.Build.props" (it makes even more sense if we later have the build tasks in ExcelDna.Build.dll)? And that means we can still add a (not user-edited) "ExcelDna.AddIn.props" later if we want to, which would have the meaning of the NuGet 2.5 props file and stay in the package directory.

I have not checked, but are the AssemblyInfo.cs files under a Properties directory for VB and F# projects too?

I have not checked, but what happens to the .props file when the package is uninstalled? I think it should remain behind...?

----

I want to think more about the rest....

caioproiete:
> OK - what do you think about the fixed name "ExcelDna.Build.props"

Works for me... PR updated with the new name.


> I have not checked, but are the AssemblyInfo.cs files under a Properties directory for VB and F# projects too?

Just checked, and they don't. VB .NET projects have a `MyProject` instead of `Properties` (with the AssemblyInfo.vb in it),  and F# projects don't have anything equivalent. The joy of Microsoft standards.

It works fine though... NuGet just creates the `Properties` folder and all works as expected. But it is weird in VB .NET projects to have it in that single folder instead of `MyProject`

I've put it back to the root of the project.


> I have not checked, but what happens to the .props file when the package is uninstalled? I think it should remain behind...?

Standard NuGet behavior (which I think makes sense): If you made any changes to the .props file, it remain behind and you get a warning saying that. If you didn't touch the .props file, then it gets removed.

If you changed the .props file, uninstalled the package (so the filed remained there), then installed again, NuGet asks you if you want to keep your file as is, or overwrite it.

govert:
I think I was confused about what you do with the different .dna files.

You understand each .dna file as creating an .xll add-in (perhaps in both bitness flavors)?

I see a project as creating a single .xll file (more precisely two version of one add-in, for the two bitness flavors). Where there are additional .dna files, these are references from the 'root' .dna file as ExternalLibraries, but don't generate separate .xll files.

I understand we might explicitly want different .dna files for the 32-bit and 64-bit outputs. By default we have one .dna file which is used for both. (I don't think of it by default as being the 32-bit specification - it's the add-in's specification file, used for both bitness flavors. That's why your names confuse me.)

Can you explain a bit how you use the idea of generating different add-ins from a single C# project?
I think these should perhaps be separate C# projects.

----

Now the behaviour I feel strongly about is this:
* Create a new C# project.
* Add ExcelDna.AddIn package.
* There is now one .dna file in the project.
* Now each of the following build customization should be easy (obvious flag name, single flag to set).
  - Don't do the packing
  - Don't build 64-bit addin
  - Don't build 32-bit addin

Beyond that I don't feel too strongly.
* Maybe if you want to make a separate add-in for every .dna file you have, then these should go into a list in the .props file (or be marked as "Build") in the project file properties.
* Otherwise it should at least be opt-out, with a way for just copying the .dna file to the output directory (as VS will do automatically if the file is marked as "Copy to Output".
* The whole thing of generating 64-bit versions from 32-bit of vice versa seems too complicated to me, but if it will help you, it can be special properties that are not in the .props by default, but can optionally be added there.

---

So here's my suggestion.

We have project-wide settings that say:
* Build 32-bit add-in(s)  -  default is 'true'
* Build 64-bit add-in(s)  -  default is 'true'

(These settings apply to all the add-ins that are build from this project.)

Now for every bitness configuration that is enabled, we need to find a .dna file. For 32-bit, we take the .dna file called xxx-AddIn32.dna, and if there is no such file, we take xxx-AddIn.dna. For 64-bit we take xxx-AddIn64.dna, and if there is no such file, take xxx-AddIn.dna. If there is no matching .dna file found, it's an error and the build fails. No flags are needed for the 'autogeneration' of anything, though the files are still copied and renamed inn the output directory according to the suffix settings.

Only unresolved issue is if you want to make different add-ins (not just different bitness) from a single C# project. I prefer to not support that for now, but it sounds like you need that support yourself. So I suggest we add an explicit list of add-ins to build in the .props file, that allows you to make many (and renamed) result add-ins, and this is where the default name of xxx-AddIn  is also overwritten.

The downside I see is that the list must be explicitly managed in the .props file, and is not implicit in the existence of .dna files. I don't know if that's a compromise you'll be OK with?

The default behaviour then adds only xxx-AddIn.dna and behaves as-if the list only has xxx-AddIn.xll in it. We might put this single entry into the .props file list, or not and have that be the implicit default.

---

I think we're just coming to these options from different perspectives. Your requirements are quite advanced, and I want to keep the simple case fairly straight-forward.

caioproiete:
Thanks @govert. ps: Do you ever sleep? :)

I'm not quite sure I understood some of the points above - I'll read it again and try to digest it, but in the meantime, maybe we can take a step back, and think about what you do today in Excel-DNA v0.32.

So, as you know better than me, this is what we get in the post-build event with a fresh install of Excel-DNA v0.32:

```
xcopy "..\packages\Excel-DNA.0.32.0\tools\ExcelDna.xll" "$(TargetDir)MyLibrary-AddIn.xll*" /C /Y

xcopy "$(TargetDir)MyLibrary-AddIn.dna*" "$(TargetDir)MyLibrary-AddIn64.dna*" /C /Y

xcopy "...\packages\Excel-DNA.0.32.0\tools\ExcelDna64.xll" "$(TargetDir)MyLibrary-AddIn64.xll*" /C /Y

"..\packages\Excel-DNA.0.32.0\tools\ExcelDnaPack.exe" "$(TargetDir)MyLibrary-AddIn.dna" /Y

"..\packages\Excel-DNA.0.32.0\tools\ExcelDnaPack.exe" "$(TargetDir)MyLibrary-AddIn64.dna" /Y
```

What I see above, is that you:

1. **Assume** that the (single) .dna file in the project will be copied to the output folder (Copy to Output)

2. **Copy** the **32-bit** .xll file with the same name of the .dna file - so, to me, you are saying that **the .dna is 32-bit by default** - completing the combination of .dna and .xll for the 32-bit flavor

3. **Clone** the .dna file to a file with the same name, but with the "64" suffix - so, to me, you are **generating a 64-bit flavor**, from the 32-bit default flavor assuming it will "just work"

4. **Copy** the **64-bit** .xll file with the same name of the newly generated 64.dna file, to complete the combination of .dna and .xll for the 64-bit flavor

5. **Pack** the 32-bit .dna, 32-bit .xll, and its dependencies (External libraries, references, etc)

6. **Pack** the 64-bit .dna, 64-bit .xll, and its dependencies (External libraries, references, etc)

---

This is the simplest scenario, and **this PR gives you this exact same behavior** since the first version, and after all the iterations we did, it continues that way.

The **only** thing I do different, is that I copy the .dna files to the output myself, instead of marking them to "Copy to Output", which I think is better / less error-prone.

Still, exact same behavior/result in the end. **Would you agree?**

---

Now, just considering those 6 steps and a fresh install, as a developer I would like be able to:

- Opt out, on packing (i.e. don't run ExcelDnaPack anything)
- Opt out, on generating the 64-bit flavor (i.e. don't create the 64.dna or copy the 64.xll)
- Make the 64-bit flavor the default, and don't generate the 32-bit flavor at all
- Make the 64-bit flavor the default, and generate the 32-bit flavor with a clone of the 64-bit .dna
- Use a different naming convention for .dna and .xll (e.g. MyLibrary-x86.dna and/or MyLibrary-x64.dna)
- Use a different suffix for the "-packed" file

---

We know we can do all this by customizing the post-builds scripts to behave the way we want, so that's fine. Moving these steps to the .targets file, however, takes control away from the developer, so we need to make sure the .props file gives use enough flexibility for these basic requirements (and I'm sure others will come up with time):

So here is how I'm resolving these, in this PR:

- [x] Opt out, on packing (i.e. don't run ExcelDnaPack at all)
Set `RunExcelDnaPack` to `false` in the .props file (it is `true` by default)

- [x] Opt out, on generating the 64-bit flavor (i.e. don't create the 64.dna nor copy the 64.xll)
Set `ExcelDnaAutoGenerate64BitAddIn` to `false` in the .props file (it is `true` by default)

- [x] Make the 64-bit flavor the default, and don't generate the 32-bit flavor at all
Rename `MyLibrary.dna` to `MyLibrary64.dna`, and set `ExcelDnaAutoGenerate32BitAddIn` to `false` in the .props file (it is `true` by default)

- [x] Make the 64-bit flavor the default, and generate the 32-bit flavor with a clone of the 64-bit .dna
Rename `MyLibrary.dna` to `MyLibrary64.dna`

- [x] Use a different naming convention for .dna and .xll (e.g. MyLibrary-x86.dna and/or MyLibrary-x64.dna)
Change the values of `ExcelDna32FileSuffix` and/or `ExcelDna64FileSuffix` in the .props file

- [x] Use a different suffix for the "-packed" file, for any flavor
Change the value of `ExcelDnaPackFileSuffix` in the .props file

---

#### After reading this. Do you still consider any of the scenarios above as "advanced"? Do you still see the same issues / confusion? Isn't it just a matter of documentation & getting acquainted with the new way of configuring things vs post-build events?

---

A few of these things impact directly on the development experience. I'll give a few (real) examples that I see with my dev team (just food for thought):

The developers in my team all use Excel 64-bit on their machines, so for building/debugging they only care about generating the 64-bit flavor as there's nothing they can do with the 32-bit anyway (i.e. avoid copying the 32-bit .xll on every build saves us a few seconds).

However, when we are ready to ship, we want to generate both flavors because some end-users have Excel 32-bit. Thus the need for flexibility - allow me to build only the 64-bit flavor during development, but build both when I compile via the build server (using MSBuild flags to override what is in the .props file). e.g. `msbuild MyProject.sln /p:ExcelDnaAutoGenerate32BitAddIn=true`

---

Packing takes time. Some of our projects have many dependencies and packing can take up to 20 seconds on every build. So we do need to have the ability to disable packing during development, but then have the ability to pack when we want to ship the add-in (via the build server). e.g. `msbuild MyProject.sln /p:RunExcelDnaPack=false`

---

Some add-ins work fine with the "unpacked" version, but do not work (crash) in the "packed" version. As an example, add-ins that use DevExpress controls, when "packed" - so for these add-ins, I need to ship unpacked - maybe this is something we can look at eventually... Not sure if it's an ExcelDnaPack problem, a DevExpress problem, or both.

---

I believe the examples above are common on any professional usage of ExcelDna... Not just at my company. Let me know if you think differently.

caioproiete:
Finally, there's one last requirement I want to mention (which I left out on purpose):

- Have separate 32-bit and 64-bit add-in specs (.dna files) for cases where each flavor needs to reference different assemblies that are platform-specific.

Maybe this is the scenario that is causing the most complication in this thread. Quoting what you said:

> I see a project as creating a single .xll file (more precisely two version of one add-in, for the two bitness flavors)

The statement above could only be true, if ExcelDna would allow me to specify all the differences about bitness in the same file. Something like this: 

```
<DnaLibrary RuntimeVersion="v4.0" > 
    <ExternalLibrary Path="Excel_XYZ.dll" /> 

    <Reference Name="MyAnyCpuLibrary.dll" /> 

    <Reference OnlyFor32Bit="true" Name="MyUSBDriver32.dll" /> 
    <Reference OnlyFor64Bit="true" Name="MyUSBDriver64.dll" /> 
</DnaLibrary> 
```

Emphasis on the (made up) `OnlyFor32Bit` `OnlyFor64Bit` attributes.

Maybe this is already possible and I just don't know how to do it - in that case, apologies for causing the confusion.

If this is not possible, then I think we need the approach I've taken in this PR, because in scenari64os like these you cannot simply copy one .dna flavor to the other.

And hence my `ExcelDnaAutoGenerate` type flag.

govert:
I'm extremely impressed and happy with your effort to move all of this into a build .targets file and I like the .props file and property-based approach to further build customization. Thank you! I also agree with almost everything you say, in particular also that your pull request version has the same effect as the old post-build events.

I'm just trying to collapse two concepts that I find confusing and redundant into a single one. I'm happy to do that myself, after merging your stuff, if that would be easier. But I want to understand your intention so that I don't inadvertently remove something you need to use.

Three of the case would be affected:
> Opt out, on generating the 64-bit flavor (i.e. don't create the 64.dna nor copy the 64.xll)
> Set ExcelDnaAutoGenerate64BitAddIn to false in the .props file (it is true by default)
I just want to call this ExcelDnaCreate64BitAddIn.

> Make the 64-bit flavor the default, and don't generate the 32-bit flavor at all
> Rename MyLibrary.dna to MyLibrary64.dna, and set ExcelDnaAutoGenerate32BitAddIn to false in the .props file (it is true by default)
I suggest this rather be addressed by setting the ExcelDnaCreate32BitAddIn to false, and *not* renaming MyLibrary.dna. Then in the build, we check according to the rules I suggested. The end result will be the same as yours.

> Make the 64-bit flavor the default, and generate the 32-bit flavor with a clone of the 64-bit .dna
> Rename MyLibrary.dna to MyLibrary64.dna
I suggest you don't need to rename the .dna file, leave it as MyLibrary.dna.

For your final scenario:
> Have separate 32-bit and 64-bit add-in specs (.dna files) for cases where each flavor needs to reference different assemblies that are platform-specific.
I suggest we need no special flags here, you just make two separate bitness-specific .dna files. According to my suggested rules for .dna file resolution that should work right.

In the light of this, it should get rid of the AutoGenerate properties, and leave us with only two properties (instead of the four):
* ExcelDnaCreate32BitAddIn
* ExcelDnaCreate64BitAddIn

The packing on/off flag, packing suffix string and bitness suffix strings stay as you have them.

----

I'm also interested in the other issue, if you have
> ModuleA.dna
> ModuleB.dna

then do you need to get separate add-ins ModuleA.xll and ModuleB.xll (and 64-bit, packed etc)?

Do you ever need to generate more add-ins (not just the 32 & 64 and -packed versions) from one C# project? How do you interpret

caioproiete:
Thank you @govert for your continuous feedback **and** patience.

Let's see if we can get this PR right in the next few days... If not I'll take you up on merging and making the changes you're thinking, and I'll test it after.

---

Number 1 is easy... Renamed.

Number 2 and 3, I *think* I understand what you mean. Let's see:

* The default platform prefix for 32-bit add-in will be "32" instead of "empty"
* The default platform prefix for 64-bit add-in continues to be "64" (no change) 
* The platform prefix can be whatever we want, as long as it is not "empty"
* For each platform, try to get the most specific file MyLibrary[Suffix].dna, fallback to MyLibrary.dna if does not exist, and fail the build if can't find MyLibrary.dna
* If you want to have control over the 32-bit and the 64-bit contents, you have to explicitly name your files MyLibrary32.dna and MyLibrary64.dna (or use whatever suffix you configured the .props file)
* Flags `ExcelDnaCreate32BitAddIn` and `ExcelDnaCreate64BitAddIn` will enable/disable the copy to the output (.dna, .xll, etc.).

Is that it? I'll have to do a couple of spikes to understand the implications of this, but indeed it looks like it invalidates the ModuleA vs ModuleB scenario where I'm generating more than one add-in per project.

I'll give this a try, but let me know if I still didn't get it.

---

Regarding ModuleA.dna and ModuleB.dna in a single project => That is correct. The expected output in this case (by default) are 4 .dna, 4 .xll, and 4 .xll-packed files, 2 per module, for 2 diff platforms.

---

> Are you OK with an explicit list in the props file, or using the .dna file properties in the project (i.e. "Build" vs. "Copy to Output") to decide this?

I think the explicit list in the .props file would be going too far - i.e. Would make it too complex to be worth it.

My 1st choice would be to have support for more than 1 add-in in the same project without anything special (drop multiple .dna files with the right suffix and it just works). If this is out of question, I'd rather live with creating a new class library per add-in, and not have to worry about MSBuild stuff anymore, other than just making small changes `true`/`false` to the .props file.
---

Well, let me take a stab at the above, and update his PR.

govert:
I _do_ want the default 32-bit suffix to stay blank. In my world, 64-bit is as useless as Mac, so I'll be switching off 64-bit immediately and won't set a suffix. In your world these are equals and you want to control the suffix.

I see one source of my confusion. You use the bitness suffix for two purposes:
* Find the right .dna file for that bitness
* Decide what the output add-in name should be

Let me try to reformulate the rules:

What 32-bit add-ins to build?
* If ExcelDnaCreate32BitAddIn is false, then none, otherwise
* Every .dna file with a 32-bit suffix, and
* Every .dna file that satisfies these two conditions:
  - it has neither the 32-bit nor the 64-bit suffix
  - there is no file with a 32-bit suffix having the same root.

What 64-bit add-ins to build?
* If ExcelDnaCreate64BitAddIn is false, then none, otherwise
* Every .dna file with a 64-bit suffix, and
* Every .dna file that satisfies these two conditions:
  - it has neither the 32-bit nor the 64-bit suffix
  - there is no file with a 64-bit suffix having the same root.

I don't know if the name check for other files is possible in the build target.

Would that catch all the cases?

---

For the multiple add-ins in one directory case, I suggest we leave it as you have it now, where every .dna file creates a set of .xlls.
Then we add an opt-out mechanism:
* If a .dna file is marked as "Copy to Output" or "Copy If Newer" in the project then it does not generate .xll files at all using your build targets.

That would give me an opt-out mechanism, should I want to add a .dna file that will only be referenced by other .dna files and is not meant to create its own .xll add-ins.

caioproiete:
Thanks @govert. I think we're close.

The rules seem clear now, however they invalidate one of the features I'd like to introduce with this PR, which is the ability to use any naming convention we want, in the source files.

In my projects, I'll have something this:

* MyLibrary-x86.dna
* MyLibrary-x64.dna

And I expect ExcelDna to understand that they are specific 32-bit and 64-bit respectively, copy the right .xll, etc., and the .dna in the output should keep the same name.

My suggestion is that we introduce properties to control source **and** destination suffixes, so that when I search for files, I can tell if they are bitness-specific or if they are "AnyCpu"

```
<!--
  Define the suffix used for each platform-specific file e.g. MyAddIn64.dna
-->
  <ExcelDna32SrcFileSuffix Condition="'$(ExcelDna32FileSrcSuffix)' == ''">32</ExcelDna32SrcFileSuffix>
  <ExcelDna64SrcFileSuffix Condition="'$(ExcelDna64FileSrcSuffix)' == ''">64</ExcelDna64SrcFileSuffix>

  <ExcelDna32DestFileSuffix Condition="'$(ExcelDna32DestFileSuffix)' == ''"></ExcelDna32DestFileSuffix>
  <ExcelDna64DestFileSuffix Condition="'$(ExcelDna64DestFileSuffix)' == ''">64</ExcelDna64DestFileSuffix>
```

Thoughts?

---

> If a .dna file is marked as "Copy to Output" or "Copy If Newer" in the project then it does not generate .xll files at all using your build targets.

I'll check if this is possible or not (I know you have to handle special cases for F# vs other types of projects, so I might have to do something similar).

However **this will break anyone using ExcelDna that have renamed their files to something other than "${projName}-AddIn.dna**. As soon as they install vNext, their .dna files will still be marked as "`Copy To Output=Copy If Newer`", and they will be ignored.

Are you OK with that, or do we need to find a different way to opt-out .xll/packing?
Could you give an example of a scenario where we want the .dna be copied to the output as is, without .xll, etc.?

govert:
Could we not have a single suffix for 32-bit, and then have MyLibrary-x86.dna build to MyLibrary-x86.xll, with the (unified) 32-bit suffix set to "-x86"?

Or do you specifically want to be able to build MyLibrary-x86.dna to MyLibrary32.xll?

I _don't_ need a scenario where (do you?):
* MyLibrary32.dna -> MyLibrary.xll
* MyLibrary64.dna -> MyLibrary64.xll


I'm happy as long as these two work:
* MyLibrary32.dna -> MyLibrary32.xll
* MyLibrary64.dna -> MyLibrary64.xll

and

* MyLibrary.dna -> MyLibrary.xll
* MyLibrary64.dna -> MyLibrary64.xll

---

I'm OK that an existing project that has renamed the .dna file and (thus also modified the post-build events) might end up in a broken state after the package upgrade. It's important that the .dna file not be deleted.

In the old UnInstall.ps1 I tried to only remove lines from the post-build script that would have been added by the Install.ps1, but that check was simplistic. So in your scenario the change .dna file and changed post-build steps would already have broken with a package upgrade.

I'm OK with that as long is no .dna file is not lost or overwritten in the process. It might be renamed.

---
I'm OK with other mechanisms to indicate that a .dna file should or should not be built. For now I at least want one way to say that a particular .dna file should not generate an .xll but should just be copied to the output. It can be with a project item setting, or by putting something into the .props file.

Maybe later I want to consider whether the extra .dna files should build a .xll by default, and whether the mechanism to build extra .xll files from extra .dna files should be different, but we need not figure that out now. I'd be more happy for now to have this work for your particular case by default, as long as there is some opt-out mechanism.

caioproiete:
Thanks @govert.

I do not need the scenario you described above. I expect the .xll to always have the exact same name as the .dna, so we're good.

---

I've done a few changes in the code this morning, and I believe it can handle all the use-cases above - will double-check and push so you can run your tests.

What it does not do, though, is handle some edge cases (likely error from the developer) which is why I'm thinking I need two prefixes (source and destination).

Thinking about something like this (not a scenario I need, but possible for a developer to create):

* MyLibrary.dna
* MyLibrary32.dna
* MyLibrary64.dna

32-bit suffix is configured to `""` (empty)
64-bit suffix is configured to `64`

Flags to create 32 and 64 add-ins all set to `true`

This would be is the result:

* MyLibrary.dna -> MyLibrary.xll (ok)
* MyLibrary32.dna -> MyLibrary32.xll (**nok**)
* MyLibrary32.dna -> MyLibrary3264.xll (**nok**)
* MyLibrary64.dna -> MyLibrary64.xll (ok)

Because the suffix for 32-bit is configured to `""` (empty), I detect MyLibrary32.dna as being "AnyCpu" and generate two versions of the add-in.

IMO this is an invalid scenario and the build should fail, but I couldn't find a way to do that, without having an indication that 32-bit source files suffix is "32". Any ideas?

Maybe I'm worrying too much?

---

Another approach (I doubt you would want that, but just in case) would be not to allow empty prefixes, set the defaults to 32 and 64, and get this:

MyLibrary.dna -> MyLibrary32.xll
MyLibrary.dna -> MyLibrary64.xll

That would allow me to fail the build reliably in the previous scenario of the 3 files. Of course, it does introduce a behavior change (i.e. the 32-bit .xll file gets the "32" suffix").

---

Did some tests and I can easily identify files marked with `Copy To Output Directory`, and the same approach works for C#, F#, VB .NET, so will move forward with that.

---

So next push should cover all use cases above +  `Copy To Output Directory` as a way to opt-out copying the .xll, packing, etc.

govert:
I would prefer to have to blank suffix option (and have that be the default for 32-bit), particularly to allow the default upgrade of an existing (unmodified) NuGet install to keep the names of the outputs the same as they are now.

I am OK with the outcome of the problem scenario you suggest - the result of:
* MyLibrary.dna -> MyLibrary.xll
* MyLibrary32.dna -> MyLibrary32.xll
* MyLibrary32.dna -> MyLibrary3264.xll
* MyLibrary64.dna -> MyLibrary64.xll

is fine in the sense that it is easy enough to explain in terms of the rules of suffix processing, even though it doesn't look right intuitively. I think it's a valid scenario, the build should not fail, and the (slightly weird) outcome is correct. [Under 64-bit Windows the 32-bit system libraries go into a directory called SysWow64. You can imagine the discussions they had...]

I suggest in the documentation (a wiki page with the fixed up version of you first description of this pull request) we describe the expected usage scenarios:

1. Default suffix settings where 32Bit="", 64-bit="64":
    (a) Single .dna file:
    * MyLibrary-AddIn.dna -> MyLibrary-AddIn.dna
    * MyLibrary-AddIn.dna -> MyLibrary-AddIn64.dna

    (b) Separate .dna files:
    * MyLibrary-AddIn.dna -> MyLibrary-AddIn.dna
    * MyLibrary-AddIn64.dna -> MyLibrary-AddIn64.dna

2. Special suffix settings where 32bit="-x86", 64-bit="-x64":
    (a) Single .dna file:
    * MyLibrary-AddIn.dna -> MyLibrary-AddIn-x86.dna
    * MyLibrary-AddIn.dna -> MyLibrary-AddIn-x64.dna

    (b) Separate .dna files:
    * MyLibrary-AddIn-86.dna -> MyLibrary-AddIn-x86.dna
    * MyLibrary-AddIn-x64.dna -> MyLibrary-AddIn-x64.dna

    (c) We make a note that this will also work:
    * MyLibrary-AddIn.dna -> MyLibrary-AddIn-x86.dna
    * MyLibrary-AddIn-x64.dna -> MyLibrary-AddIn-x64.dna

Any other scenarios can go wrong as far as I'm concerned, as long as we can describe the processing rules.

caioproiete:
Great @govert. I'll update this PR at some point this week after testing all the use-cases described above and collate the contents of the many comments above, in a wiki page for a first draft and we go from there.

govert:
@caioproiete Wonderful - thank you!

There is no rush. I really appreciate your effort.

jiripik:
The current script modifies the debug target, adds a blank dna file and modifies the Post Build events even in a project with the DNA already set up.

I do understand that the automatic installation into a non-blank project is a challenge but I would recommend doing in that case nothing but a string replace of the directory name with the old version with the new version. Anything else messes up the current project.

konne:
@caioproiete did you already updadet the PR?

konne:
@caioproiete any news, I really like to change my csproj to the new build targets from you

caioproiete:
Hi @konne, no not yet.

Manual tests were taking too much time, so I decided to write integration tests for everything, which got me a little bit further, but the MSBuild code to support all scenarios was getting more and more complicated, and I ended up with a C# MSBuild task and wrote unit tests that I can easily debug, etc.

This is where I am at the moment:
image


##### Default suffix settings where 32Bit="", 64-bit="64":

- [x] . (a) Single .dna file:
MyLibrary-AddIn.dna -> MyLibrary-AddIn.dna
MyLibrary-AddIn.dna -> MyLibrary-AddIn64.dna

- [x] . (b) Separate .dna files:
MyLibrary-AddIn.dna -> MyLibrary-AddIn.dna
MyLibrary-AddIn64.dna -> MyLibrary-AddIn64.dna


##### Special suffix settings where 32bit="-x86", 64-bit="-x64":

- [x] . (a) Single .dna file:
MyLibrary-AddIn.dna -> MyLibrary-AddIn-x86.dna
MyLibrary-AddIn.dna -> MyLibrary-AddIn-x64.dna

- [x] . (b) Separate .dna files:
MyLibrary-AddIn-86.dna -> MyLibrary-AddIn-x86.dna
MyLibrary-AddIn-x64.dna -> MyLibrary-AddIn-x64.dna

- [ ] . (c) We make a note that this will also work:
MyLibrary-AddIn.dna -> MyLibrary-AddIn-x86.dna
MyLibrary-AddIn-x64.dna -> MyLibrary-AddIn-x64.dna


I'm not sure yet how to support the last scenario without complicating things too much.

Will try to push the code I have over the weekend.

govert:
@caioproiete That's amazing. We now have better testing for the build script than for any other part of Excel-DNA!

To be clear - I don't really have an opinion about what should happen in that slightly awkward outstanding case (c). I just meant that we just need to make a note of whatever happens for that corner case (even if the outcome is bad in some sense).

konne:
@caioproiete That's cool

konne:
@caioproiete: any news. If you make a fork and put your code in I can also support you, but I don't see any progress since months

caioproiete:
Hi @govert / @konne / @jiripik

Sorry for the huge delay. Life got in the way and took me some time to be able to come back to this one. Better later than never (?) :).

I've just updated the pull-request with what I believe to be a v1.0 of a full MSBuild-based pipeline using custom tasks.

Pretty much all the the features, behaviors, and whatnot that we discussed above are still valid and should "just work".

I'll let you digest it and (hopefully) take it for a spin and try to break it. The integration tests cover (I think) all the important scenarios, so I'm not expecting too many bugs, but you never know, right? :)

---

To the solution, I've added two projects:
- **Source\ExcelDna.AddIn.Tasks\ExcelDna.AddIn.Tasks.csproj**
- **Source\Tests\ExcelDna.AddIn.Tasks.IntegrationTests\ExcelDna.AddIn.Tasks.IntegrationTests.csproj**

---

#### Source\ExcelDna.AddIn.Tasks\ExcelDna.AddIn.Tasks.csproj
This project implements the custom MSBuild task and its output (ExcelDna.AddIn.Tasks.dll and ExcelDna.AddIn.Tasks.pdb) are now parte of the ExcelDna.AddIn NuGet package (in the tools folder) along with the `ExcelDna.AddIn.targets` file which does the orchestration in a very simplified way (you will notice that all the previously complicated MSBuild code in this file have been removed and the C# task is doing all the hard work).

**Package\ExcelDna.AddIn\tools**
image


#### Source\Tests\ExcelDna.AddIn.Tasks.IntegrationTests\ExcelDna.AddIn.Tasks.IntegrationTests.csproj
This project implements lots of integration tests to exercise / validate (most) of the scenarios supported. The tests are written with [NUnit](http://nunit.org/) and can be run from the IDE if you have ReSharper or similar, otherwise you'll need to download the [NUnit runner](http://nunit.org/index.php?p=download). There's an additional piece of work to be done, which is setup these tests to be run in the AppVeyor build (TBD).

image


These integration tests rely on a separate set of project files (and its own solution) under **Tests\ExcelDna.AddIn.Tasks.IntegrationTests.TestTarget**. This separation is required so that we can execute MSBuild separate from Visual Studio and not have files locked, etc. The integration tests essentially call MSBuild via command-line, targeting each of these projects - which are simple Excel-DNA add-ins pre-configured to execute with the same MSBuild pipeline that is used by the ExcelDna.AddIn NuGet package.

image


---

#### Expected behavior after installing the ExcelDna.AddIn NuGet package

- Post-build events are no longer added to the project (yay!). The only thing added to the project is a reference to the `ExcelDna.AddIn.Targets` file, in the tools folder of the NuGet pacakge. If upgrading the package in an existing project, the uninstaller of the previous version will attempt to clean-up the build events, so in most cases it should be a clean upgrade (assuming no big customization was made by the user);

- By default, the .dna file is copied to the build output in two flavors, 32-bit and 64-bit;

- By default, the .xll file is copied to the build output in two flavors, 32-bit and 64-bit;

- By default, the ExcelDnaPack is executed for each .dna file created by the build process in the output;

- By default, if an App.config file exists, it is copied to the build output for each .xll created by the build process (with the same name of the xll + ".config"). This is also true for .xll files created by ExcelDnaPack (i.e. the external .config file corresponding to the -packed.xll file is always made available if the source .xll had a config file); It's possible to have specific config files per platform (32/64-bit) and there is a fallback to App.config.

- All the behavior above can be customized by the user via the file `Properties\ExcelDna.Build.Props` or via command-line MSBuild parameters (e.g. from your continuous integration server). The user can enable/disable the creation of platform-specific output, change naming conventions, disable or customize packing (disable compression or multi-threading).

---

#### Known issues / Possible things for a future release

- There is one edge case scenario which is not currently supported, when the output .dna file must have a suffix that is different than the suffix used in the input .dna file:
`MyLibrary-AddIn.dna` -> `MyLibrary-AddIn-x86.dna`
`MyLibrary-AddIn-x64.dna` -> `MyLibrary-AddIn-x64.dna`

- There is pretty much no support for incremental builds, which means that the files .dna, .xll, .xll.config, etc. are always copied to the output, regardless if they have changed - this is no different than what the `xcopy` command does today in the post-build events, by the way so not really an issue. I'm just pointing this out as a possible area for improvement in a future release (i.e. check last write date and/or size of files before copying).

- When copying the files to the build output folder, in case something goes wrong (e.g. file is locked by a process and cannot be overwritten) the build will fail right away. There is no support for N retries, w/ pauses, etc.

- When copying the files to the build output folder, in case the destination file is marked as read-only, the copy will fail. In the future, we should check for the read-only attribute first and remove it, in order to be able to overwrite the file.

- I think (in the future) it would be a good idea to make ExcelDnaPack a proper MSBuild task, instead of just spawning a process via Exec in the MSBuild file. Once this PR gets in and settles down, that could be a good move.

---

Looking forward to getting feedback on this PR.

govert:
@caioproiete Thank you very much for this!'
It looks like you've done a great job.
I'll go through it and then see how to pull it in for the next version.

.gitattributes
```diff
@@ -4,6 +4,11 @@
 * text=auto
 
 ###############################################################################
+# Explicitly declare files that should always be converted to LF regardless of platform
+###############################################################################
+*.dotsettings text eol=lf
```
govert: Need to review what this means.

Build/build.bat
```diff
@@ -3,3 +3,5 @@ copy /Y ..\Source\ExcelDna\Release64\ExcelDna64.xll ..\Distribution\
 copy /Y ..\Source\ExcelDna.Integration\bin\Release\ExcelDna.Integration.dll ..\Distribution\
 copy /Y ..\Source\ExcelDnaPack\bin\Release\ExcelDnaPack.exe ..\Distribution\
 copy /Y ..\Source\ExcelDnaPack\bin\Release\ExcelDnaPack.exe.config ..\Distribution\
+copy /Y ..\Source\ExcelDna.AddIn.Tasks\bin\Release\ExcelDna.AddIn.Tasks.dll ..\Package\ExcelDna.AddIn\tools\
+copy /Y ..\Source\ExcelDna.AddIn.Tasks\bin\Release\ExcelDna.AddIn.Tasks.pdb ..\Package\ExcelDna.AddIn\tools\
```
govert: Is there a particular reason to have the .pdb in the NuGet package?
(In general I'd prefer a smaller package).
Is there some reference to current recommendations on this?
Do we need to add the ExcelDna.Integration.pdb too?

Package/ExcelDna.AddIn/content/ExcelDna.Build.props
```diff
@@ -0,0 +1,47 @@
...
+
+  <!--
+    Configuration properties for packing .dna files
+  -->
+  <PropertyGroup>
+    <!--
+      Enable/Disable packing of .dna files
+    -->
+    <RunExcelDnaPack Condition="'$(RunExcelDnaPack)' == ''">true</RunExcelDnaPack>
+
```
govert: In future we can add the NoCompression packing options (which speeds up the builds).

Package/ExcelDna.AddIn/content/ExcelDna.Build.props
```diff
@@ -0,0 +1,47 @@
...
+    <!--
+      Suffix used for packed .xll files e.g. MyAddIn-packed.xll
+    -->
+    <ExcelDnaPackXllSuffix Condition="'$(ExcelDnaPackXllSuffix)' == ''">-packed</ExcelDnaPackXllSuffix>
```
govert: What happens if it's blank? (overwrite the .xll?)
Does this give us a way to put the packed files in a separate directory?

Source/ExcelDna.sln
```diff
@@ -1,6 +1,8 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2012
+# Visual Studio 14
```
govert: Does anything depend on this change?

.gitattributes
caioproiete: `.dotsettings` files are the files where we can configure ReSharper settings for the solution. These files should always have Unix line-endings, even on Windows, so this is what this does... Make sure `.dotsettings` files are always converted to LF when you clone the repository / checkout a branch.

Build/build.bat
caioproiete: It's a common practice to include .pdb files in NuGet packages, so that we can get detailed exception information, in the case of errors.

ExcelDna.Integration.pdb is not required for this PR to work, but I would recommend it to be included in future packages for the same reason.

Most people nowadays don't commit nuget packages to source control, and use package restore, so saving one or two megabytes in the package size makes little difference.

Package/ExcelDna.AddIn/content/ExcelDna.Build.props
caioproiete: The NoCompression flag is already supported... [See the .targets file](https://github.com/Excel-DNA/ExcelDna/pull/27/files/764ce5d26c7dc272cf83c52e696dfaf376eeddda#diff-62f2797be89bc72359a89542376edfbdR117). I just didn't include it in the .props file, thinking that this is more of an advanced option.

i.e. if the developer adds the following to the .props file, it will work:

`<ExcelDnaPackCompressResources Condition="'$(ExcelDnaPackCompressResources)' == ''">false</ExcelDnaPackCompressResources>`

The same applies to the NoMultithreading packing option

`<ExcelDnaPackRunMultithreaded Condition="'$(ExcelDnaPackRunMultithreaded)' == ''">false</ExcelDnaPackRunMultithreaded>`

Package/ExcelDna.AddIn/content/ExcelDna.Build.props
caioproiete: Good question. I didn't test this. It will do whatever `ExcelDnaPack.exe` decides to do... I'd expect it to override the original XLL, but it's worth testing it. In case it doesn't override, we could add validation to make sure this is never empty.

Source/ExcelDna.sln
caioproiete: Nothing really "depends" on this change for it all to work, but this makes the solution compatible with Visual Studio 2015, whilst keeping compatibility to previous version - i.e. `MinimumVisualStudioVersion = 10.0.40219.1` so it's harmless.

Package/ExcelDna.AddIn/content/ExcelDna.Build.props
govert: Maybe it would be useful to have something like `<ExcelDnaPackXllDirectory>`?
(Perhaps future version).
The point is that one would typically want to test the packed file in a directory of its own, to see that the dependencies are actually packed, and not loaded from files.

Package/ExcelDna.AddIn/content/ExcelDna.Build.props
```diff
@@ -0,0 +1,47 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003" DefaultTargets="ExcelDnaProps">
+<!--
+  If you change properties in this file, they may not come into effect until you:
+     * Rebuild the solution/project
+
+     or
+
+     * Close Visual Studio
+     * Delete .vs folder, if exists
```
govert: Really? Is this a known bug or something?

Package/ExcelDna.AddIn/content/ExcelDna.Build.props
govert: OK cool - I agree they are advanced options, so we don't put them in the .props file by default.

Package/ExcelDna.AddIn/content/ExcelDna.Build.props
caioproiete: Yes, that seems like a good idea.

Package/ExcelDna.AddIn/content/ExcelDna.Build.props
caioproiete: It's a known issue with some versions of Visual Studio and for some types of projects, which seem to cache the .targets (or .props) files and not automatically reload it after you make changes/save. In my tests with VS2015 and C# projects, I've never experienced this, but I did see this problem with F# projects a few months ago when I first sent the pull-request, and after some Googling, I thought it would be better to write this notice here.

PR #34, opened 2015-09-21
* add /NC option (no Compression) for packing without lzma

konne:
if you have big projects the packing can take a long time, for testing if packing is working fine the version without LZMA is much faster (like my project 40sec. -> 3sec.)

There is a new commad line parameter /NC -> this means no compression

Signed-off-by: Konrad Mattheis <konrad.mattheis@akquinet.de>

### DIFF
--- a/Source/ExcelDnaPack/ExcelDnaPack.csproj
+++ b/Source/ExcelDnaPack/ExcelDnaPack.csproj
@@ -28,6 +28,8 @@
     <WarningLevel>4</WarningLevel>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <LangVersion>3</LangVersion>
+    <PlatformTarget>x86</PlatformTarget>
+    <UseVSHostingProcess>true</UseVSHostingProcess>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <DebugType>pdbonly</DebugType>
--- a/Source/ExcelDnaPack/PackProgram.cs
+++ b/Source/ExcelDnaPack/PackProgram.cs
@@ -22,6 +22,7 @@ Usage: ExcelDnaPack.exe dnaPath [/O outputPath] [/Y]
 
   dnaPath      The path to the primary .dna file for the ExcelDna add-in.
   /Y           If the output .xll exists, overwrite without prompting.
+  /NC          no compress (LZMA) of resources
   /O outPath   Output path - default is <dnaPath>-packed.xll.
 
 Example: ExcelDnaPack.exe MyAddins\FirstAddin.dna
@@ -82,6 +83,7 @@ Other assemblies are packed if marked with Pack=""true"" in the .dna file.
             string xllInputPath = Path.Combine(dnaDirectory, dnaFilePrefix + ".xll");
 			string xllOutputPath = Path.Combine(dnaDirectory, dnaFilePrefix + "-packed.xll");
             bool overwrite = false;
+            bool compress = true;
 
 			if (!File.Exists(dnaPath))
 			{
@@ -107,6 +109,10 @@ Other assemblies are packed if marked with Pack=""true"" in the .dna file.
                     else if (args[i].Equals("/Y", StringComparison.CurrentCultureIgnoreCase))
                     {
                         overwrite = true;
+                    } else
+                    if (args[i].Equals("/NC", StringComparison.CurrentCultureIgnoreCase))
+                    {
+                        compress = false;
                     }
                 }
             }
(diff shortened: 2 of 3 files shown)


konne:
Hi govert,

I also added the mutlithreading for the lzma compression. This reduce the packing time for my project to the half.

please check and give me a feedback.

govert:
Hi Konrad,

I'm happy with the suggestion, but will have to go through the code changes.
Could you maybe suggest a different name for the flag? I'd prefer something more descriptive than /NC

-Govert

konne:
Hi Govert

- flag Ideas: /no_lzma /nolzma, better?

-SevenZipHelper: yes, because the compression properties are static defined, I already testing this changes sinces about 3month with no issue.

bye
Konrad

govert:
What about /NoCompression ?
(I'm also tempted to suggest /NoDepression)

If we manage to refactor the build process a bit more, as started in #27  this could become an option in the ExcelDna.Build.props file.

---
I'm a bit unsure about using the enum .ToString() values directly.
Could you at least add some comments to the enum that indicates that the values are used this way, and that the names cannot be changed.

Also, where AddFile is called with the compression option always set to false (e.g. for .config files) we should note that this cannot be changed without changing the usage inside the native code or ExcelDna.Loader.

govert:
Multithreaded compression:

Would it be a good idea to keep all the resource updates on the same thread (they're serialized with a lock anyway) and just have the compression be done on the worker threads?
That might require a bit more bookkeeping for the threadpool tasks, but would get rid of the worker threads running in other parts of the ResourceHelper code.

I'm also a bit by the non-determinism we're introducing by the update sequence not always being the same. Is there a way to get the resource update sequence always be the same, even though the compression tasks might run in parallel?

konne:
>What about /NoCompression ?
okay, I change it.

>BuildTasks, I have seen this, and this is also a later task for me to make a real msbuildtask out of the ExcelDNAPack, so you don't have to call it via <Exec...>

>I'm a bit unsure about using the enum .ToString() values directly,...
I will comment this and have the plan to move this at a general level. I used already the same "ids / numbers" what you are using in the .Loader. I suggest to change "internal static byte[] GetResourceBytes@AssemblyManager.cs", the type = 0. If the enum is general in a later step this code can also change.

konne:
>Multithreaded compression
It is really faster if we update the resource in a separate thread. Even if it has to wait sometimes a little bit, because of the lock. 
The sequence of updating the resources has no effect. For all resources that are accessed by "name" I see nothing. The only point is that typelib resources can have a different typelib number.
I can change that point, but the code is than more complex and need more memory, because we have to make a byet[] -> list of all compressed resources and update this after all compression is done.

konne:
FYI, I also working on a bigger change (http://exceldna.codeplex.com/discussions/632029#post1416974) but with the real usage of the build in localisation feature of FindResourceEx,... and not the first working but not so nice approach from my codeplex discussion.

govert:
OK - we'll try it with the multi-threaded resource updates, and if there issues we can add a switch to do a deterministic single-threading pack.

govert:
Can you explain about the TypeLib resources?

konne:
every time you add a Typelib, you increase the typelibnumber and add each typelib only with a number as reference. 
>ResourceHelper.UpdateResource(_hUpdate, typeName, (IntPtr)typelibIndex, localeNeutral, pinHandle.AddrOfPinnedObject(), (uint)data.Length);

So It could be happen that typelibs get different numbers, because of different calling of the adding.
But I never saw that anybody was complaining that he has to change the ordering of the references in the DNA file.

konne:
Hi Govert,

I have changed the NC to /NoCompression.
Could you already test the multithreading of lzma, did you find any problems?

bye
Konrad

govert:
@konne I have not tried it yet.

I'll check it, and if all looks OK, I'll accept to pull request.
I might then add a new Issue for us to consider maybe adding another flag that forces a deterministic (single-threaded build). The issue would serve as a reminder that this is something to look out for, if anyone else is bothered by the new behaviour.

konne:
No Problem, I will add another Command Line Parameter /NoMultiThreaded okay for the naming?
Stopping the Multithreading is very easy :-)

govert:
OK cool - that would be great. I guess "NoMultiThreading" is better than "NoMultiThreaded".

konne:
Features is added, please test and add to the mainline

govert:
@konne I've changed the CLA handler to CLA assistant since CLA Hub is not working anymore.
I can't really see what the experience is from your perspective.

You should have seen a prompt when updating the other PR. If there's a problem, please let me know.

Otherwise, I'm happy to accept this PR - in my (fairly quick) tests it worked fine.

konne:
I got an CLA email and I accepted the rules

konne:
it is stranged I accept the cla and the system says, thank you for accepting and than the cla is still not okay

govert:
@konne I can see that you agreed to the CLA. Not sure why it didn't pick it up.

Thank you very much for your contribution to making Excel-DNA better!

PR #35, opened 2015-09-26
Use locales

konne:
This update fixes the not clean handling of Sattelite Resource Assemblies. Till that time is was not possible to add Sattelite Assemblies, because they have alle the same name and only a different culture.
The following Picture shows the packing of with the new ExcelDnaPack. You can see clear that the CSHARPADDIN.RESOURCES are added with 2 different Languages.

packed_satteliteassemblies

This Commit also add all the Changes to the ExcelDna.Loader/AssemblyManager.cs to reflect the loading of different Cultures. It also includes a Fallback, so if no fitting Culture is found it try to loads the default (first added) resource without culture.

In the samples I added a small Sample to test the new functionality with an easy test.

Please test this very soon because I like to get this tested and approved very fast.


govert:
@konne You'll have to be a bit more patient with me on this feature, as you're making changes to the add-in runtime (ExcelDna.Integration) and not just the build process as in #34.

konne:
@govert: I made even more rework of the AssemblyManager.cs that take care of all Assemblies, if you have any question just contact me.
The complete handling is no change from the simple name to the fullname (including version, culture,..)

konne:
@govert 
>By checking against that full name of the loaded..
That is not right, I only check the fullname for all already loaded assemblies, the check against the packed file is the same with name in uppercase. So it can never happen that a load that was working before can fail.

What really can happen in a special case is that a already loaded assembly is loaded twice. But I don't see a real case when it can happen, because I try to do my best to don't load it twice like you (check against assemblies in Domain / check against own dictionary,...)
Even if I don't believe that its happen, I can will add a hash over the byte and store them in a separate dictionary, than it can never happens that the same bytes[] are loaded twice. Okay?

govert:
I'll have to think about an example to explain the loading issue.
I'm not sure a byte[] hash makes it better - it just adds complication to a very sensitive part of the add-in runtime.

I'm just not yet comfortable with the changes we're making to the add-in runtime here. You'll have to be a bit more patient with me for this PR.

Source/ExcelDna.Loader/AssemblyManager.cs
```diff
@@ -36,77 +38,92 @@ internal static void Initialize(IntPtr hModule, string pathXll)
...
-				loadedAssembly = Assembly.Load(assemblyBytes);
-                Logger.Initialization.Info("Assembly Loaded from bytes. FullName: {0}", loadedAssembly.FullName);
-				loadedAssemblies.Add(name, loadedAssembly);
-				return loadedAssembly;
-			}
-			catch (Exception e)
-			{
+            //File.WriteAllBytes(@"c:\Temp\" + name + ".dll", assemblyBytes);
+
+            // compute the hash of the bytes
```
konne: @govert this codeline prevent 100% against loading more than one time from the packed resources, because it checks the real loaded bytes

konne:
no problem, I know that this is a very sensitive part of the hole plugin. I also added a sample VS Project in my branch. So you can play with it a little bit.
Just take the time, to test and understand my approach.

Bytes-Hash: I would integrate this, because this makes it 100% safe. I already made a commit for it, this are only about 10 line of code more and they are not complex.

govert:
I don't think I like using the assembly hashes there - for one it would not protect against mistakenly loading an assembly via the AssemblyResolve that was loaded in another Load Context. The name check would probably prevent this, but I the whole thing just makes me nervous.

konne:
I understand now you problem, we can make the namecheck without the the version, so it only loads assemblies with different cultures. This checking can also made against the AppDomain so it get the info over all loaded Assemblies.
What we can do is also to write into the Resource File also the Fullname of an example, so we don't have to load the assembly to get which assembly is in the resource file. What do you think about this?

konne:
cla is now fine

konne:
@govert I changed now the checking fur already loaded assemblies to Name+Culture, so if you really have assemblies with different version, it will not load the assembly again

govert:
@konne Thank you. I should be able to have a look in the last week of October.

konne:
@govert Do you already had the time to look into that release.

govert:
@konne I will try over the next few days to sort it out, but have not had another look yet. Could you perhaps check why the AppVeyor build is failing with the current check-in? I think it is to do with the sample project you added.

konne:
@govert I will fix this on the weekend, yes it is just a absolute path in the sample project

govert:
OK - just to update where I stand on this.

I'm happy for us to add support for localized resource assemblies to the Excel-DNA packing.
One possible future use of this would be to have localized function descriptions etc.

The implementation of this needs to make minimal changes to the existing assembly loading code. In my experience this code is quite fragile - the assembly resolution from the .NET runtime is very poorly documented and Excel-DNA already goes quite far in interfering with the normal loading. In practice this means that it must be clear to me, by inspection, that any changes would not interfere with the existing loading.

I think the current pull request changes the assembly loading implementation too much for me to be comfortable.

Some of the complexity in the current PR is because our AssemblyLoad implementation also does the fallback to the Neutral culture if the required assembly is not found. From this it seems the mechanism of storing and checking hashes arises.

Is this the right place to implement the missing culture fallback?

It seems the current implementation would behave differently to unpacked .NET assemblies, in that a failing request for a specific culture would fall back to the current thread culture, rather than the neutral culture.

I'd also like to know how NeutralResourcesLanguageAttribute is used by .NET (via the AssembyLoad event).

I think I need to play a bit more, to see what I would be happy with. I might like to keep the changes you have in the ExcelDnaPack, but reduce the changes to the AssemblyLoad event.

konne:
Hi govert,

here some comments on your long comment:
>One possible future use of this would be to have localized function descriptions etc.
Yes, I will introduce this in my plugin with dynamic register of excel function, but it would be much nicer to have an attribute that just point on a resource.
This also is very nice for alle WPF / WinForms Dialogs or CustomsToolBars, because all the localisation is working with this changes out of the box.

>Change AssemblyLoad
I know that this is a tricky part and I really tried over a long time not to change this code.

>Is this the right place to implement the missing culture fallback?
I have just added the fallback to have the same behavior like your implementation. So if somebody uses an old dnapack, or don't use the localisation it works in the old behavior.

>It seems the current implementation would behave differently to unpacked .NET assemblies
Yes, but this is already the actual situation. So I know that my solution doen't solve all issues, but it already helps a lot.

> NeutralResourcesLanguageAttribute
I will test this, but I think this should already work out of the box, because this resources are embedded already in the dll and not in the sattelite assemblies. So this language is already loaded before it tries to find the language resource in the sattelite dlls.

>From this it seems the mechanism of storing and checking hashes arises.
>+ .. changes you have in the ExcelDnaPack
I really like to change the ExcelDNAPack and also the packing of the resource. Why?
It would be great to get the fullname of an loaded assembly (bytes from resource) before I load this with LoadAssembly. It would be also possible to use somethink like mono.cecil, but I think it would be much better to add a header to each resource there the AssemblyFullname with Culture,.. is readable for the LoadAssembly function.

I'm happy to get more feedback from you and I also like to change the code in a way that you are more comfortable with.

bye
Konrad

govert:
I have some small questions about one piece of the code in `AssemblyManager.AssemblyResolve`:

```
            // if we don't have a culture info or of it is the Neutral culture take 0, otherwise take LCID
            if (assName.CultureInfo != null &&  !string.IsNullOrEmpty(assName.CultureInfo.Name))
                lcid = (ushort)assName.CultureInfo.TextInfo.LCID;

            // 4096 -> 0x1000 are user defined LCIDs / Cultures
            if (lcid == 4096)
                return null;
```

1. Is there any reason to use AsesmblyName.CultureInfo.TextInfo.LCID` instead of `AssemblyName.CultureInfo.LCID`?
2. Do you know of any case where AssemblyName.CultureInfo` is not null, but AssemblyInfo.CultureInfo.LCID` would not be valid? (Do we need the `IsNullOrEmpty` check?)
3. Why would we explicitly not support cultures with LCID > 4096, if the assebmly has that LCID?

konne:
>AsesmblyName.CultureInfo.TextInfo.LCID

because this LCIDs are like the old W32Api IDs, if you open the XLL in a ResourceExplorer you see the correct language to each resource

>string.IsNullOrEmpty(assName.CultureInfo.Name))

This is empty "" for the neutral culture and for old Win32Api I like to have the LCID=0

>Why would we explicitly not support cultures with LCID > 4096

This only blogs the =4096 and not greater than 4096. The Problem is that you can have 10 assemblies with user defined cultures and this can not saved into on 4096 LCID, because they alld have the same id.
If we like to support that, we have to kick out the old Win32Api LCIDs and save it in a different way into the packed XLL.

bye
Konrad

govert:
Are you saying that `AssemblyName.CultureInfo.TestInfo.LCID` is different to `AssemblyName.CultureInfo.LCID`?

Are you saying that `AssemblyName.CultureInfo.LCID` is not `=0` for the Neutral culture?

Would you suggest that instead of storing the culture in the Windows PE resource culture, we rather just encode it in the name, e.g. MYASSEBMLY.RESOURCES_1023

konne:
>Are you saying that AssemblyName.CultureInfo.TextInfo.LCID is different to ...

in some cases

>Are you saying that AssemblyName.CultureInfo.LCID is not =0 for the Neutral culture?

yes

>Would you suggest that instead of storing the culture in the Windows PE resource culture, we rather just encode it in the name, e.g. MYASSEBMLY.RESOURCES_1023

I don't like that approach. (may be MYASSEMBLY.RESROUCES_EN_US)
 What I like is that we introduce into the XLL file a real header with pointers to all emedded assemblies. And that all resources are just numbers.

govert:
I'd be OK with something like `MYASSEMBLY.RESOURCES_EN_US` instead of storing it in the Windows PE resource identifier. Is there any downside or advantage beyond the user-defined cultures to doing  this?

I don't _yet_ see any value in changing the general packing story. If you think it has benefit, I suggest you make a separate issue where we can discuss that. The localization seems simple enough to pin down on its own, without having to think about the packing layout etc. now.

konne:
>Is there any downside or advantage beyond the user-defined cultures to doing this

the advantage at the moment is, that its clear separated (name and lcid) and not mixes in one string. 
There can be sometimes sideeffects if for example somebody has an assembly like: ass1_en.dll which has nothing todo with the language english.

>I don't yet see any value in changing the general packing story

I don't like to make a separate thread for it, because this discussion is very close to the LoadAssembly discussion. It would be much easier in LoadAssembly if the function can read out all Assemblies/Names/Cultures out of the XLL file without really loading them with Assembly.Load.

govert:
One confusion when using the Windows PE Resource language is that FindResourceEx will match a fallback language (i.e. will return the resource with language en-US when asked for en). If your code asks for "en-ZA" and that fails, .NET will ask for "en" only, but if AssemblyResolve returns the "en-US" assembly, .NET still won't use it. 

Seemingly it would be easier to match .NET's interpretation if the language is just encoded in the resource name string. If we only do this for ".resources" assemblies, it seems the chance of an unintended problem is small.

I see that it simplifies a lot of other things. I'll look more closely in the next few days - @konne maybe you can wait a little bit before working more on this.

konne:
>  If we only do this for ".resources" assemblies

yes

>maybe you can wait a little bit before working more on this

okay, just contact me and than we can discuss what are the best next steps

govert:
@konne I've now made another pull request #48 which implements the localization with only adjusted names, as we discussed here. This changes the code a lot less.

I'm hoping it will satisfy your requirements. Perhaps you can have a look to check that I haven't forgotten anything.

I've also added a LocalizedResources sample in the Samples repository (https://github.com/Excel-DNA/Samples/tree/master/LocalizedResources) as a basis for a sample that show how the feature would work. For now one needs to manually patch the packages directory with the new files. Later it should work with the appropriate package version.

konne:
The changes from @govert in PR #48 also solves that issue quite well.

PR #43, opened 2015-11-05
Fixes #14 when args array is empty

### LINKED ISSUE #14: Fix NuGet install script for F# in Visual Studio 2015
When installing the NuGet package into an F# project in VS 2015, there is an error:
```
Exception setting "Value": "Object of type 'Microsoft.VisualStudio.FSharp.ProjectSystem.BuildAction' cannot be converted to type 
'VSLangProj.prjBuildAction'."
At C:\Temp\Fs2015Test\packages\ExcelDna.AddIn.0.33.7-rc1\tools\install.ps1:56 char:13
+             $newDnaFile.Properties.Item("BuildAction").Value = ([Microsoft.Visua ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], SetValueInvocationException
    + FullyQualifiedErrorId : ExceptionWhenSetting
```

This is not a serious setting to change, and everything else still seems to work.

Another F#-related problem in the NuGet install script is to configure the Debug executable. F# uses a different project system and this setting is not (easily) accessible.

tomikiss:
Updated LogDisplay

### DIFF
--- a/Source/ExcelDna.Integration/LogDisplay.cs
+++ b/Source/ExcelDna.Integration/LogDisplay.cs
@@ -264,7 +264,7 @@ namespace ExcelDna.Logging
         {
             lock (SyncRoot)
             {
-                string message = string.Format(format, args);
+                string message = args.Length > 0 ? string.Format(format, args) : format;
                 string[] messageLines = message.Split(LineEndChars, StringSplitOptions.RemoveEmptyEntries);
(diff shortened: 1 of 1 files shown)


govert:
Cool - thank you!

Issue #45, opened 2015-11-06
Last argument description of every function has ". " added to it, even if it already ends that way.


alainbryden:
Minor issue worth resolving in the next release.

The code [here](https://github.com/Excel-DNA/ExcelDna/blob/faf87765376a51fb06b9ae795d88a6652c31a5e5/Source/ExcelDna.Loader/XlRegistration.cs#L282) meant to fix the "Excel Function Description bug" is resulting in already well formatted function argument descriptions getting two periods:

![untitled](https://cloud.githubusercontent.com/assets/2285037/11001454/a8cc69b0-847d-11e5-95f5-caf59a26116f.png)

The solution is probably as simple as trimming off any trailing period and space before adding it back:

```
argumentDescriptions[j] = argumentDescriptions[j].TrimEnd(new[] { ' ', '.' }) + ". ";
```

This will play nice with people whose argument descriptions already end in ". " or just ".";

---

If we want to be more sensitive to people who choose to have things like ellipses at the end of their descriptions (and we probably do), maybe something more long winded, like:

```
if (!argumentDescriptions[j].EndsWith(". "))
{
    if (argumentDescriptions[j].EndsWith("."))
        argumentDescriptions[j] += " ";
    else
        argumentDescriptions[j] += ". ";
}
```

That way descriptions like:
`"Custom tags, metadata, etc..."` will be converted to:
`"Custom tags, metadata, etc... "` rather than
`"Custom tags, metadata, etc.... "` or
`"Custom tags, metadata, etc. "`

caioproiete:
LGTM.

@alainbryden: Will you send a pull-request?

PR #46, opened 2015-11-09
Fix for issue #45

alainbryden:
Proposed fix for https://github.com/Excel-DNA/ExcelDna/issues/45#issuecomment-154597865


govert:
Cool @alainbryden  - thank you!
Should we do the same for functionDescription, as few lines down?

alainbryden:
Hmm, yes I think so. I had ignored it because at a glance I thought it only applied when the description was empty, but it looks like it will run when there are no function arguments but the description is non-empty.

alainbryden:
Pushed and squashed additional fix.

govert:
Thanks Alain!

PR #48, opened 2015-11-12
Implement packing for localized .resources assemblies

govert:
This is an attempt at a simplified version of the localization support discussed in #35 .
Instead of using the PE file support for adding a culture to a resource, we just append the name of the culture to the packed resource name (and check this when doing the AssemblyResolve).

### DIFF
--- a/Source/ExcelDnaPack/ResourceHelper.cs
+++ b/Source/ExcelDnaPack/ResourceHelper.cs
@@ -178,7 +178,7 @@ internal unsafe static class ResourceHelper
         {
             lock (lockResource)
             {
-                bool result = ResourceHelper.UpdateResource(_hUpdate, typeName, name, localeEnglishUS, IntPtr.Zero, 0);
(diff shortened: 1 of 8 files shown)


Source/ExcelDnaPack/ResourceHelper.cs
```diff
@@ -167,7 +178,7 @@ public void RemoveResource(string typeName, string name)
         {
             lock (lockResource)
             {
-                bool result = ResourceHelper.UpdateResource(_hUpdate, typeName, name, localeEnglishUS, IntPtr.Zero, 0);
+                bool result = ResourceHelper.UpdateResource(_hUpdate, typeName, name, localeNeutral, IntPtr.Zero, 0);
```
konne: nice that you also fixes this. I had already seen this, but didn't want to add another building lot

konne:
You changes work quite well in my big excel plugin, so It believe that it will also work for all others.
thx konne

govert:
OK - I've pulled these changes into master.

PR #52, opened 2015-11-23
initialize the LogStrings

dreamershl:
It can avoid the NULL exception issue when LogDisplay is configured to be top trace listener. Since it is static, instead of checking the NULL in the function, I create the instance when the class is constructed. So no need NULL check anymore.


govert:
Great - thank you very much!

PR #53, opened 2015-11-30
Set accessibility properties

andrewkittredge:
This changes are required to make Add-ins with the log Section 508 compliant, necessary if you want to sell to the United States government.

http://www.section508.gov/


govert:
Great - thank you!

Issue #63, opened 2016-03-09
Multithreaded packing fails for large number of assemblies


govert:
When packing a large number of assemblies (about 80), the new multi-threaded packing fails with a message

```
System.NotSupportedException: The number of WaitHandles must be less than or equal to 64.
   at System.Threading.WaitHandle.WaitAll(WaitHandle[] waitHandles, Int32 millisecondsTimeout, Boolean exitContext)
   at System.Threading.WaitHandle.WaitAll(WaitHandle[] waitHandles)
   at ResourceHelper.ResourceUpdater.EndUpdate(Boolean discard)
   at ResourceHelper.ResourceUpdater.EndUpdate()
   at ExcelDnaPack.PackProgram.Pack(String[] args)
   at ExcelDnaPack.PackProgram.Main(String[] args)
```

There is a limit on the number of events we can wait for, and currently we make an event per assembly that is packed.

@konne If you're still around, would you perhaps like to look at this?

A workaround is to run the ExcelDnaPack without multi-threading, by adding the `/NoMultiThreading` switch to the command line.

konne:
I will change the code so if the packing is more than 64 the rest is packed not multithreaded.
I try to fix it next week.

govert:
@konne - Thank you!
Alternatively we just keep a (concurrent) queue of assemblies being compressed, remove assemblies as they are done, and signal the event when the queue is empty. The only little snag is that we're still targeting .NET 2.0, so need to do the locking ourselves (no concurrent collections yet).

konne:
@govert 
I sent you an pull request. Hope that it fixes this for ever :-)

govert:
Done - thank you @konne !

PR #66, opened 2016-03-11
fix issues #63 for multithreaded packing of more than 64 items

konne:
the Endupdate checks now always if 64 Thread are done and after this it checks the next 64 Threads, so it should work fine with large amounts of packages

Signed-off-by: Konrad Mattheis <konrad.mattheis@akquinet.de>


govert:
@konne Great - thank you!

PR #78, opened 2016-05-04
Remove ExcelAsyncUtil obsolete method from samples

kchen0723:
Remove ExcelAsyncUtil obsolete method from samples


govert:
Thanks - that's great. I've been neglecting the samples a bit.

Issue #79, opened 2016-05-20
make IsMainThread from internal to public


kchen0723:
The reason is that in some cases, we need to call excel com object in the excel main UI thread. so we need to check whether the thread is the main thread. So we need to expose this property

PR #80, opened 2016-05-20
expose IsMainThread property

kchen0723:
To fix the issue:
https://github.com/Excel-DNA/ExcelDna/issues/79

### DIFF
--- a/Source/ExcelDna.Integration/Excel.cs
+++ b/Source/ExcelDna.Integration/Excel.cs
@@ -62,11 +62,27 @@ namespace ExcelDna.Integration
                 _mainWindowHandle = GetWindowHandleThread();
         }
 
-        internal static bool IsMainThread
+        public static bool IsMainThread
         {
             get
             {
                 return Thread.CurrentThread.ManagedThreadId == _mainManagedThreadId;
+            }
+        }
+
+        public static int MainManagedThreadId
+        {
+            get
+            {
+                return _mainManagedThreadId;
+            }
+        }
+
+        public static long MainNativeThreadId
+        {
+            get
+            {
+                return _mainNativeThreadId; //conver to long so that it is CLSCompliant.
             }
         }
 


govert:
In practice:
* The main thread id is always 1.
* The .NET runtime makes no guarantee that the native thread id is stable - it might change, so I don't want to expose it with an API
* Knowing that you are running on the main thread (`IsMainThread == true`) does not ensure that you can safely interact with Excel through COM or the C API.

I suggest for this pull request that we only make the IsMainThread proparty public, but not add the extra accessors for the Managed Thread and Native Thread Ids.

govert:
@kchen0723 I'm kind of reconsidering whether it would be a good idea for ExcelDna to expose these as public members.

1. You can trivially implement it yourself - the main thread will have managed thread id 1.
2. The native thread Id is unreliable and using it is deprecated in .NET. So it seems inappropriate to expose it. The add-in can recover it from the managed thread id as well.
3. Exposing IsMainThread as a property might give the add-in developer the false idea that either the C API or the COM object model can safely be called. This would not be true in general.

Given these doubts, maybe it's better to not make those members part of the public interface at the moment. What do you think?

kchen0723:
@govert I spent 2 days to investigate the main thread id issue. Yes, you are right, the main thread Id is always 1(why? I still do not know). 

Also the IExcelAddIn.AutoOpen() is called by DnaLibrary.AutoOpen(), which is called by ExcelIntegration.DnaLibraryAutoOpen(), This one will called by ExcelDna.cpp.xlAutoOpen. So I think the IExcelAddIn.AutoOpen() is called in the Excel Main UI thread, right? Thanks.

govert:
Yes - it's guaranteed that AutoOpen is called on the main thread. It's also a safe place to make COM calls, e.g to hook up Application events.

On 27 May 2016, at 19:18, kchen0723 <notifications@github.com<mailto:notifications@github.com>> wrote:


@govert<https://github.com/govert> I spent 2 days to investigate the main thread id issue. Yes, you are right, the main thread Id is always 1(why? I still do not know).

Also the IExcelAddIn.AutoOpen() is called by DnaLibrary.AutoOpen(), which is called by ExcelIntegration.DnaLibraryAutoOpen(), This one will called by ExcelDna.cpp.xlAutoOpen. So I think the IExcelAddIn.AutoOpen() is called in the Excel Main UI thread, right? Thanks.

kchen0723:
Ok, I get it. 
Since the AutoOpen is called on the main thread,We can do our logic here. 
So we need not this change. I will close this pull request. Thanks so much.

PR #99, opened 2016-12-02
A test case to run XL test workbooks

masariello:


govert:
Instead of adding the vs2015 C++ project and solution, could you copy the old one and rename to .vs2013. instead. Then I'm happy for the 'current' project to use the VS2015 toolset.
I'd prefer to leave the output from the main project as "ExcelDna.xll" and "ExcelDna64.xll".

masariello:
Yep. Will do.

masariello:
github seems to have a problem with the way I set up the commits. I will close this request and open a new one from a new hopefully correct branch.

PR #110, opened 2017-01-27
Fix #109 Excel-DNA build fails with read-only package

### LINKED ISSUE #109: Excel-DNA build fails with read-only package
Tracking this issue: https://groups.google.com/d/topic/exceldna/S9ucYrB6k_g/discussion

caioproiete:
This PR fixes issue with not replacing read-only files in the build output


govert:
@caioproiete Oh cool - thank you!

PR #113, opened 2017-02-14
ExcelDnaPack: Set Environment.ExitCode to 1 by default (gets set to 0 (zero) when success)

### LINKED ISSUE #112: ExcelDnaPack should return a failure code if the specifed .dna file is not found

caioproiete:
Fixes #112 (?)


govert:
Great, thanks @caioproiete  - I think that should work.

PR #114, opened 2017-02-17
add facility for copying external library version to xll version

rosecodym:
I had a similar problem to this: https://groups.google.com/forum/#!topic/exceldna/HlVeZUZNcqM Basically, I wanted the .xll generated by ExcelDnaPack to have a version number other than the one from the template .xll. VerPatch (linked from that discussion) worked, but actually figuring out how to integrate it into my workflow was tricky. Patching ExcelDnaPack turned out to be easier.

You can now mark ExternalLibrary elements in the .dna file with UseVersionAsOutputVersion="true" to copy their version resource into the output .xll. If you specify this for more than one ExternalLibrary only the first will be used, and a warning will be emitted. If the version copying fails for some reason (like the source file doesn't have it, e.g. if it's a .dna file), then a warning is emitted.

This seemed like the most straightforward way to do this, but I'm totally open to suggestions for refinements or things I've missed.


caioproiete:
Wow! This looks great! Thank you @rosecodym!

govert:
@rosecodym Cool, I've merged it in - thank you very much!

PR #125, opened 2017-03-26
Change to build clean task

sizzles:
Changes for the build clean task to remove the delete * which was causing problems when building multiple add-ins to the same directory.

Also fixed some integration tests that didn't seem to be working.
I did this by adding the following to the csproj files.

  <PropertyGroup>
    <PreBuildEvent>ECHO Running ExcelDna **Pre-build** event</PreBuildEvent>
  </PropertyGroup>
  <PropertyGroup>
    <PostBuildEvent>ECHO Running ExcelDna **Post-build** event</PostBuildEvent>
  </PropertyGroup>

I've kept the methods / names as close as possible to the existing.
So there is probably some room for tidy up around that (particularly in relation to packed add-in deletion). Very happy for feedback etc etc

Also the issue of cleaning add-in files that get renamed (currently they will not be cleaned as there is no tracking of them. The delete * did not have this problem as everything was cleaned up)

### DIFF
--- a/Package/ExcelDna.AddIn/tools/ExcelDna.AddIn.targets
+++ b/Package/ExcelDna.AddIn/tools/ExcelDna.AddIn.targets
@@ -1,5 +1,6 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <UsingTask TaskName="ExcelDna.AddIn.Tasks.CreateExcelAddIn" AssemblyFile="ExcelDna.AddIn.Tasks.dll" />
+  <UsingTask TaskName="ExcelDna.AddIn.Tasks.CleanExcelAddIn" AssemblyFile="ExcelDna.AddIn.Tasks.dll" />
 
   <PropertyGroup>
     <ExcelDnaTargetsImported>true</ExcelDnaTargetsImported>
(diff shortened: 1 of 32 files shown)


caioproiete:
Nice! Thanks a lot @sizzles!!

Gave it a cursory look just now, and looks great. Will pull your branch and give a more thorough look tomorrow and give feedback.

I'm surprised that you had to add `ECHO Running ExcelDna **Pre-build** event` etc. to several projects. The only place I'm asserting this is on a single test in ([PreAndPostBuildEventsIntegrationTests](https://github.com/Excel-DNA/ExcelDna/blob/master/Source/Tests/ExcelDna.AddIn.Tasks.IntegrationTests/PreAndPostBuildEventsIntegrationTests.cs#L20-L23)) which only targets the project `PreAndPostBuildEvents`, so I'll look at this tomorrow as well.

ps: Would be great if you could `rebase` / clean-up the branch and `push --force` as looks messy with all the merge commits you had to bring, etc.

sizzles:
@caioproiete  - Yes, for tests I cloned in a new repo and could reproduce what was happening.
Should have some time next couple of days to look at rebasing / tidying up.

sizzles:
@caioproiete - As per your suggestion, I've re-based and squashed my commits, so this history should be cleaner now.  Hope it looks ok.  

Thanks

Issue #119, opened 2017-03-13
Feature request: Class / Module Custom Attributes


ittegrat:
It could be useful to introduce some class / module custom attributes to specify registration properties for a group of functions / commands.

Something like:
```
<ExcelFunctionGroup(Category:="MyCat", IsHidden:=true)>
Public Module MyModule
  <ExcelFunction(Name:="MyAdd")>
  Function funcA(x, y)
    return x + y
  End Function   
  <ExcelFunction(Name:="MyMult, IsHidden:=false")>
  Function funcB(x, y)
    return x * y
  End Function
End Module
```
where the function / command attributes take precedence over the group attributes for the specified fields.

Thanks !

ittegrat:
Not of general interest.
Use Registration extension instead.

PR #131, opened 2017-05-15
Add support for group ExcelFunctionAttribute and ExcelCommandAttribute

ittegrat:
ExcelFunctionAttribute and ExcelCommandAttribute can be now declared at
class / Module level and they are applied to the contained methods (cfr. #119).
Attributes declared at method level take precedence over the ones specified
at class level for what they declare (cfr. Distribution/Samples/GroupAttributes.dna
for examples).

### DIFF
--- a/Distribution/Samples/ExplicitExports.dna
+++ b/Distribution/Samples/ExplicitExports.dna
@@ -23,6 +23,28 @@
 		End Function	
 	End Module
 
+	<ExcelFunction(Category:="ExplicitModule")> _
+	Public Module VisibleModule
+		Function MultiplyThem(x, y)
+			MultiplyThem = x * y
+		End Function
+
+    <ExcelFunction(Name:="SubtractThem")> _
+		Function SubThem(x, y)
+			SubThem = x - y
+		End Function
+	End Module
+
+	Public Module InvisibleModule
+		Function MultiplyThem(x, y)
+			MultiplyThem = x * y
+		End Function
+
+		Function SubThem(x, y)
+			SubThem = x - y
+		End Function
+	End Module
+
 ]]>
 </Project>
 </DnaLibrary>
--- a/Distribution/Samples/GroupAttributes.dna
+++ b/Distribution/Samples/GroupAttributes.dna
@@ -1,0 +1,45 @@
+<DnaLibrary Name="Attributes Sample" Language="C#">
+<![CDATA[
+
+  using ExcelDna.Integration;
+
+  [ExcelFunction(Category="VolatileFuncs", IsVolatile=true)]
+  public class MyClass {
+
+    public static double funcA(double x, double y) {
+      return (x + y) ;
+    }
+
+    [ExcelFunction(IsVolatile=false)]
+    public static double funcB(double x, double y) {
+      return (x * y) ;
+    }
+
+    [ExcelFunction(Name="Cfunc")]
+    public static double funcC(double x, double y) {
+      return (x - y) ;
+    }
+
+    [ExcelFunction(Category="NotVolatileFuncs", IsVolatile=false)]
+    public static double funcD(double x, double y) {
+      return (x / y) ;
+    }
+
+  }
+
+  [ExcelCommand(MenuName="SampleMenu")]
+  public class MyMenuClass {
+
+    [ExcelCommand(MenuText="MenuItem 1")]
+    public static void aaa() {}
+
+    [ExcelCommand(MenuText="MenuItem 2")]
+    public static void bbb() {}
+
+    [ExcelCommand(MenuName="SampleMenu 2", MenuText="Item 1")]
+    public static void ccc() {}
+
+  }
+
+]]>
+</DnaLibrary>
--- a/Distribution/Samples/GroupAttributesVB.dna
+++ b/Distribution/Samples/GroupAttributesVB.dna
@@ -1,0 +1,46 @@
+<DnaLibrary Name="AttributesVB Sample" Language="VB">
+<![CDATA[
+
+  <ExcelFunction(Category:="VolatileFuncs", IsVolatile:=true)> _
+  Public Module MyModule
+
+    Function funcA(x As Object, y As Object) As Object
+      return x + y
+    End Function
+
+    <ExcelFunction(IsVolatile:=false)> _
+    Function funcB(x As Object, y As Object) As Object
+      return x * y
+    End Function
+
+    <ExcelFunction(Name:="Cfunc")> _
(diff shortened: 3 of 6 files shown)


govert:
@ittegrat Thank you for the pull request. I still need to have a closer look.

I wonder how this interacts with the explicit exports option - are all functions in a class with an ExcelFunction attribute always exported?

I wonder if this can be done with the Registration extensions, without changing the base Excel-DNA library. For this I don't know if derived attribute classes can change the places where the attribute is applicable.

ittegrat:
@govert Thank you for your attention. I've strived to keep changes backward compatible and as minimal as possible.

I've updated the PR to handle the `ExplicitExports` attribute. `Samples/ExplicitExports.dna` shows the extended behavior.

ittegrat:
If I didn't misinterpreted your code, the actual logic is:
```
  for each candidate method
    if it is marked with an ExcelAttribute
      record it
    else
      record a null as a placeholder
```
I simply extended this logic:
 ```
  for each candidate method
    if it is marked with an ExcelAttribute
      if the containing type is marked with a matching ExcelAttribute, merge infos
      record it
    else
      if the containing type is marked with an ExcelAttribute
        record it
      else
        record a null as a placeholder
```
benefiting of all the already implemented logic to select the candidate methods and to check if the method is a supported one.

As far as I can understand the Registration library, I should implement again all the selection of the candidate methods. Am I wrong ?

**Comment by govert** [comment]
I want to look at this a bit more carefully - adding the larger scope for the attribute must integrate with the various selection flags, and with the listing that is done in the Registration extension. I also think the benefit is quite small (essentially only the Category and IsHidden attributes can be usefully re-used). So I'm not going to bring it in for the current version v0.34, which I hope to release as it is in the RC now.
Then I'll have a look again and see how best to incorporate your proposal for the next version.

**Comment by caioproiete** [comment]
I really like the idea of having attributes that can be applied to an entire class for things like setting the category of functions.

Instead of changing the existing `ExcelFunctionAttribute` and `ExcelCommandAttribute` classes, I'd suggest we create specific attributes for each class-level setting, such as:

* `ExcelFunctionCategoryAttribute` - Sets a category name for all functions in a class
* `ExcelFunctionIsHiddenAttribute` - Sets the "is hidden" flag for all functions in a class
* `ExcelCommandMenuNameAttribute` - Sets the menu name for all commands in a class
* etc.

These attributes should only be allowed at a class level, and they would be overridden if the user specifies their values in `ExcelFunction` or `ExcelCommand` for any particular function/command.

e.g.

    [ExcelFunctionCategory("Blue")]
    public class MyClass
    {
        // Will be added to the "Blue" category, because of the class-level setting
        [ExcelFunction()]
        public static double FunctionA(double x, double y)
        {
            return (x + y) ;
        }
    
        // Will be added to the "Red" category, because user is overriding class-level category
        [ExcelFunction(Category = "Red")]
        public static double FunctionB(double x, double y)
        {
            return (x * y) ;
        }
    }

**Comment by govert** [comment]
I'm still not sure about this. It seems to add little value and can be done in the Registration extension anyway. How many hidden functions do people register? How much does the category matter now that we have the in-sheet IntelliSense? If every function gets an attribute anyway, how hard is adding the Category?

Note that this pattern, where a custom add-in is derived from ExcelFunction, works with the current version. So you need not put the attribute on the class, just use your own attribute with the defaults baked in.

```
public class MySpecialExcelFunctionAttribute : ExcelFunctionAttribute
{
    public MySpecialExcelFunctionAttribute() 
    {    
        Category = "My Special Functions";
    }
}

public static class MyFunctions
{
    [MySpecialExcelFunction(Name="Cool.Function")]
    public static double MyFunction(double arg1) {...}
}
```
Maybe that's a bit nicer.

On the other hand, I don't see much harm to allowing the class-level attribute, beyond possible adding some quirks.

Even then, I don't really like the `ExcelFunctionCategoryAttribute` - to me it looks like even more confusion, even though I understand it makes the class-level settings more explicit.

Let me start by some specific issues:
* If a class is marked as `ExcelFunction` but the assembly is ExplicitExports=true, does it mean the functions themselves still need to be marked in order to show?
* Why stop at the class level, can't the whole assembly have an ExcelFunction attribute?
* How hard will it be to make the Registration extension consistent - at least the helper that does the method enumeration and pulls out the attributes?

Overall I'm not super excited about this, but if you want to do it let's just pin down the details.

caioproiete:
Hey @ittegrat I'll continue the conversation, but feel free to chime in... Not trying to steal your thunder!

This is definitely not a high priority, but rather a nice-to-have, and looks simple enough to implement.

---

> can be done in the Registration extension anyway

True. Although I think very few people use the Registration extension (at least in my experience). The attributes that come out-of-the-box are usually good enough for most cases, and the first examples devs bump into all use the attributes.

---

> How many hidden functions do people register?

Probably not many. Personally I'm more interested in the category class-level definition which is the only thing in this PR I'd really use, tbh.

---

> How much does the category matter now that we have the in-sheet IntelliSense?

I think it matters quite a bit, specially when you have different teams building add-ins and bringing in dozens of functions in each. The categories (when used properly) can help users understand which add-in is providing which function, as well as making it easier to search/find the right functions when you know the add-in that provides it.

The Intellisense is great, but I was not able to roll it out company wide yet. It's still not stable enough (as of June 2017 at least, didn't try again since then) and it started crashing user's Excel quite frequently. I'll eventually try again and try to collect some logs to see if it helps resolving this issue.

---

> If every function gets an attribute anyway, how hard is adding the Category?

Definitely not hard. Just a little annoying having to repeat the same thing on every function.

---

> If a class is marked as ExcelFunction but the assembly is ExplicitExports=true, does it mean the functions themselves still need to be marked in order to show?

Good question. I'd say no... The functions all should inherit the class-level attribute, and therefore they are automatically all exported by inheritance.

However, I believe I'm thinking this way because you're talking about putting an `ExcelFunction` attribute in the class definition. If the name of the attribute was different (e.g. just `Category` from `System.ComponentModel`), I would think every function inside the class would need an `ExcelFunction` in order to be exported, if `ExplicitExports=true`.

---

> Why stop at the class level, can't the whole assembly have an ExcelFunction attribute?

No reason... Could be assembly level too. Class-level & assembly-level. With the class-level overriding assembly-level, and function-level overriding class-level.

---

> How hard will it be to make the Registration extension consistent - at least the helper that does the method enumeration and pulls out the attributes?

Doesn't look like a big job, but I haven't touched the extensions in a while, and could be forgetting something. But I get your point that we should merge this PR together with a PR to the extensions and have them aligned.

PR #134, opened 2017-07-31
Update GettingStarted.txt directions

autobotsassemble:
Seems like RuntimeVersion is case sensitive, should be = "v4.0" not "V4.0"


govert:
That's true - thanks!

Issue #158, opened 2018-01-03
Support dynamic help address resolution


imanushin:
For current moment developer have to put url to help topic into ExcelFunction attribute. So, it is unable to change URL dynamically, therefore it is unable to select different urls for production and staging environments.

What is expected: instead of http address. put link to static method, which will resolve help page dynamically.

E.g. signature can be:

```
public static Url ResolveHelpPage(ExcelFunctionAttribute targetMethod)
```

caioproiete:
@imanushin This can be done with the [Excel-DNA Registration Helper](https://github.com/Excel-DNA/Registration), a library that allows you to dynamically register your functions.

[Install the package `ExcelDna.Registration` in your project](https://www.nuget.org/packages/ExcelDna.Registration/), so you can have access to the custom registration classes.

[![image](https://user-images.githubusercontent.com/177608/34572493-e362dd78-f147-11e7-9b8c-cb918a8a7acb.png)](https://www.nuget.org/packages/ExcelDna.Registration/)

In your .dna file, set the flag `ExplicitRegistration` to `true` to tell Excel-DNA that you'll be doing the registrations of the functions yourself.

e.g.
```xml
<DnaLibrary Name="MyAddIn Add-In" RuntimeVersion="v4.0">
  <ExternalLibrary Path="MyAddIn.dll" ExplicitRegistration="true" (...) />
</DnaLibrary>
```

Declare your functions normally:
```csharp
public static class MyFunctions
{
    [ExcelFunction]
    public static string Hello()
    {
        return "Hi";
    }

    [ExcelFunction]
    public static string Hello2()
    {
        return "Hi2";
    }
}
```

In your `AutoOpen`, register the functions, replacing the `HelpTopic` of each of the functions as you need.

e.g.
```csharp
public class MyAddIn : IExcelAddIn
{
    public void AutoOpen()
    {
        RegisterFunctions();
    }

    public void AutoClose()
    {
    }

    private static void RegisterFunctions()
    {
        var registrationEntries =
            from assembly in ExcelIntegration.GetExportedAssemblies()
            from type in assembly.GetTypes()
            from methodInfo in type.GetMethods(BindingFlags.Public | BindingFlags.Static)
            where methodInfo.GetCustomAttribute<ExcelFunctionAttribute>() != null
            select new ExcelFunctionRegistration(methodInfo);

        var delList = new List<Delegate>();
        var attList = new List<object>();
        var argAttList = new List<List<object>>();

[...]

PR #159, opened 2018-01-08
Convert fields in our attributes to virtual properties, to allow dynamic behavior

caioproiete:
This PR makes it easier to implement dynamic behavior in our attributes such as described in #158.

By allowing devs to create their own `ExcelXXXAttribute`-derived classes and overriding its properties, we can more easily allow for dynamic behavior in said properties, without requiring fallback to custom registration.

For example, (after this PR gets merged) to implement a dynamic `HelpTopic` as described in #158, developers can create their own `ExcelFunctionAttribute`-derived class and build the `HelpTopic` dynamically:

```csharp
[AttributeUsage(AttributeTargets.Method, Inherited = false, AllowMultiple = false)]
public class ExcelFunctionV2Attribute : ExcelFunctionAttribute
{
    public override string HelpTopic
    {
        get
        {
            var environment = ConfigurationManager.AppSettings["Environment"];
            return $"http://{environment}/help/{Name}";
        }

        set { throw new NotImplementedException(); }
    }
}
```

And, of course, decorating the function methods with the derived class, for it to work:

```csharp
public static class MyFunctions
{
    [ExcelFunctionV2]
    public static string Hello()
    {
        return "Hi";
    }
}
```

### DIFF
--- a/Source/ExcelDna.Integration/ExcelAttributes.cs
+++ b/Source/ExcelDna.Integration/ExcelAttributes.cs
@@ -42,9 +42,17 @@ namespace ExcelDna.Integration
 	[AttributeUsage(AttributeTargets.Parameter, Inherited = false, AllowMultiple = false)]
 	public class ExcelArgumentAttribute : Attribute
 	{
-		public string Name = null;
-		public string Description = null;
-		public bool   AllowReference = false;
+		private string _description;
+
+		public virtual string Name { get; set; }
+
+		public virtual string Description
+		{
+			get { return _description; }
+			set { _description = value; }
+		}
+
+		public virtual bool AllowReference { get; set; }
 
 		public ExcelArgumentAttribute()
 		{
@@ -52,7 +60,7 @@ namespace ExcelDna.Integration
 
 		public ExcelArgumentAttribute(string description)
 		{
-			Description = description;
+			_description = description;
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


govert:
As is, this will break the current reflection code that reads the attribute information, but that code can be changed fairly easily too.

But I'm not sure the field->property change is worth the risks. As you've pointed out in the initial issue, this case can be dealt with using the registration helper. I'd rather have some helper or better documentation for the registration stuff, than fiddle too much on the core Excel-DNA internals.

caioproiete:
I'm happy to update the PR with fixes to the reflection code if you decide to move forward with this one. Let me know.

Is it just [here](https://github.com/Excel-DNA/ExcelDna/blob/f67a391857700f3cec4855c7c361cec27a709a4f/Source/ExcelDna.Loader/XlMethodInfo.cs#L207), or somewhere else? Also, any reason not to just do a cast to the appropriate class, and not have to use reflection at all to read the fields/properties?

Either way, I agree we need to improve the registration story to make it easier... The amount of code I had to "borrow" from the registration library, in the example for #158 is not ideal for sure.

govert:
@caioproiete Let's wait a bit on changing the attribute stuff to properties.

The reason for using reflection is that there might be multiple instances of the ExcelDna.Integration assembly loaded into memory into different .NET 'load contexts'. This causes the type from the add-in assembly and the type used in the ExcelDna.Loader to be considered different by the runtime (there's no type unification across load contexts, and then it all fails. It's a mess in the code, but works for now so I have little incentive for the effort to address this.

The code you show in #158 might be simpler is you called `ExcelRegistration.GetExcelFunctions()`, then modify, then call `.RegisterFunctions();` like in the sample add-in here https://github.com/Excel-DNA/Registration/blob/master/Source/Samples/Registration.Sample/ExampleAddIn.cs . But it's not documented anywhere, which I think is the main problem. And generally feels a bit user hostile. I'm not sure how to think about it.

caioproiete:
Closing this one as `wont-implement`. Recommended work-around is to use **[`ExcelDna.Registration`](https://github.com/Excel-DNA/Registration)** to register the functions explicitly at runtime.

PR #164, opened 2018-01-14
Bugfix #156 decorate excel dna with resharper attributes

### LINKED ISSUE #156: Decorate Excel-DNA plugin with ReSharper attributes
To help ReShaper analyze code better, it is better to decorate API with set of special attributes:
1. ```ExcelFunctionAttribute``` can be decorated as [MeansImplicitUseAttribute](https://www.jetbrains.com/help/resharper/Reference__Code_Annotation_Attributes.html#MeansImplicitUseAttribute). So, ReShaper will understand, that method, which was decorated by ```ExcelFunctionAttribute```, is called from Reflection API
2. XlCall methods can be decorated with NotNull/CanBeNull attributes (to show warnings in the compile time)
3. There are different ways, how to reference attributes:
3.1 Add dependency and exclude it from nuget package. Code will work fine even if end-used will not have this library in the AppDomain. However, ReSharper will understand dependencies
3.1 Generate internal implementation (e.g. with internal visibility, [file will be like this](https://github.com/imanushin/CheckContracts/blob/master/CheckContracts/Properties/JetBrainsAttributes.cs)) ReSharper annotations from MSVS and add them into final dlls. So, ReShaper will able to read internal method (and do appropriate highlighting), however end-user will not see them.

imanushin:
Fixes #156 issue

### DIFF
--- a/Source/ExcelDna.Integration/Excel.cs
+++ b/Source/ExcelDna.Integration/Excel.cs
@@ -12,7 +12,7 @@ using System.Reflection;
 
 namespace ExcelDna.Integration
 {
-    public class ExcelDnaUtil
+    public static class ExcelDnaUtil
     {
         private delegate bool EnumWindowsCallback(IntPtr hwnd, /*ref*/ IntPtr param);
         private delegate bool EnumThreadWindowsCallback(IntPtr hwnd, /*ref*/ IntPtr param);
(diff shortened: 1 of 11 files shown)


govert:
OK great - I don't use Resharper myself, but the changes don't look dangerous.

PR #167, opened 2018-02-15
Update NuGet package to use `build` feature for adding `.targets` files to projects

caioproiete:
This PR improves the installation process of our `ExcelDna.AddIn` NuGet package by moving the `.targets` file to a `build` folder, [which tells NuGet to (automatically) include the targets file to the project file](https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package#including-msbuild-props-and-targets-in-a-package) as part of installing the NuGet package.

With that feature, we are able to remove most of the Powershell code that we currently have on `install.ps1` and `uninstall.ps1` to handle just that adding and removing of the `.targets` file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ Package/nupkg/
 
 # Enable "build/" folder in the NuGet Packages folder since NuGet packages use it for MSBuild targets
 !packages/*/build/
+!Package/*/build/
 
 # MSTest test Results
 [Tt]est[Rr]esult*/
--- a/Distribution/ChangeLog.txt
+++ b/Distribution/ChangeLog.txt
@@ -2,6 +2,7 @@ Version 0.35 (???)
 ------------------
 * Improve build tasks - more reliable clean-up and debugger detection.
 * Improve RTD and async QueueAsMacro reliability
+* Improve install process of ExcelDna.AddIn NuGet package (now requires NuGet 2.5)
 
 Version 0.34 (18 June 2017)
 ---------------------------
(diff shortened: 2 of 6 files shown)


govert:
How does this feature interact with the minimum NuGet version required?
Does it mean installation will now fail under older Visual Studio versions?
If so, I think that is OK, but we should try to figure out something to say in the readme file for such users.

caioproiete:
> How does this feature interact with the minimum NuGet version required?
> Does it mean installation will now fail under older Visual Studio versions?

If the user has an older version of NuGet (older than 2.5 in this case) the package will not install and the user will get an error message. This is described in the [NuGet documentation](https://docs.microsoft.com/en-us/nuget/release-notes/nuget-2.5#add-a-minimum-nuget-version-property-to-packages).

The version of Visual Studio is not really important AFAIK. Only the NuGet extension needs to be (relatively) up-to-date - i.e. `>= 2.5`, to use this feature.

NuGet 2.5 was released on April 25, **2013**, so devs had plenty of time to update :wink:

---

> ... we should try to figure out something to say in the readme file for such users.

Updating the readme of the NuGet package doesn't really help, because the readme only opens after the package installs - and if the NuGet client is old the package won't even install, and no chance of reading the readme.

I've updated the changelog to reflect this new requirement, and once the next version of the NuGet package gets published to NuGet.org, it will automatically display on the description of the package that we require NuGet 2.5 or higher (because of the `minClientVersion` attribute).

Something like this:

image

PR #194, opened 2018-05-19
Add IntelliSense to .dna files inside Visual Studio

### LINKED ISSUE #28: Create an .xsd file for the .dna files
Creating an .xsd schema file for the .dna files should lead to an improved editing experience in Visual Studio.

caioproiete:
This PR adds an `.xsd` file that describes the schema of `.dna` files, which Visual Studio can leverage to provide the IntelliSense to `.dna` files as well as perform validation checks within the XML editor. Resolves #28.

image

image

image

Sample project to see a working example:
* [Awesome.zip](https://github.com/Excel-DNA/ExcelDna/files/2019282/Awesome.zip)
  ( _just open the solution in Visual Studio and start playing with the file `Awesome-AddIn.dna`_ )

### How it works

The `ExcelDna.AddIn` nuget package now includes a file called `ExcelDna_dna_2018_05.xsd` which gets added to the project, and Visual Studio can see it.

The schema declares an XML namespace (that I've made up) called `http://schemas.excel-dna.net/addin/2018/05/dna.xsd`

The `ExcelDna-Template.dna` that comes in the `ExcelDna.AddIn` nuget package now includes that same namespace `http://schemas.excel-dna.net/addin/2018/05/dna.xsd` - which is the hint that Visual Studio needs to understand that it should use the schema `ExcelDna_dna_2018_05.xsd`  for IntelliSense on the `$(ProjectName)-AddIn.dna` file.

This means that:
1. If you are creating a brand new project and installing `ExcelDna.AddIn` for the first time, everything "just works" and you get IntelliSense on the `.dna` file immediately.

2. If you are upgrading `ExcelDna.AddIn` on an existing project, you'll need to manually add the namespace `http://schemas.excel-dna.net/addin/2018/05/dna.xsd` to your `.dna` file in order to get the IntelliSense.

```xml
<DnaLibrary Name="MyAddIn" (...) xmlns="http://schemas.excel-dna.net/addin/2018/05/dna.xsd">
  <ExternalLibrary Path="Awesome.dll"  ExplicitExports="false" (...) />
  <!-- ... -->
</DnaLibrary>
```

#### Elements
I've added IntelliSense to all elements possible in a `.dna` file, starting with `DnaLibrary`, and including `ExternalLibrary`, `Reference`, `Project`, `SourceItem`, and `Image`. With the exception of `CustomUI` which has its own schema defined by Microsoft and is not part of Excel-DNA.

image

#### Attributes
I've added IntelliSense to all attributes in each element above such as `RuntimeVersion`, `Language`, `ExplicitExports`, etc. with the exception of `CompilerVersion` which currently is not validated by the Excel-DNA Add-In loader.

image

I've tried to design the XSD to be as permissive as the C# code that validates the `.dna` files is, which means accepting elements to be in any random order as well as accepting different options for attributes. For example, `Language` that can be `CSharp` or `C#` or `CS`, or `F Sharp` or `fsharp` etc... And Visual Studio will not display validation errors for these different possibilities as they are all valid in the C# code.

However, the IntelliSense auto-complete will only show the "preferred" suggestions, which in the example of `Language` would be only three: `C#`, `VB`, and `F#`

#### Compatibility
I've tested this XSD against all `.dna` files in the [`/Distribution/Samples/`](https://github.com/Excel-DNA/ExcelDna/tree/master/Distribution/Samples) folder as well as with all the `.dna` files in the [`Samples`](https://github.com/Excel-DNA/Samples) repository, and they all pass validation :heavy_check_mark:, including the ones that contain custom code as pure text in `CDATA` regions as well as the ones that have custom `Ribbon`s and other `CustomUI` elements.

### Schema versioning

Versioning is done through the XSD file name (ExcelDna_dna_**`2018_05`**.xsd) and through the XML namespace (http\://schemas.excel-dna.net/addin/**`2018/05`**/dna.xsd). Whenever we make changes to the `.dna` file format, we can update the schema accordingly and give it a new file name and XML namespace.

### Validation of .dna files during build time

Visual Studio automatically performs validation checks **in the editor** when we open the `.dna` file and shows visual cues and warnings to the developer. based on the `.xsd` that I'm introducing with this PR However this PR does not implement this validation during the build (`ExcelDnaBuild` task) which means the build will succeed regardless of having a valid `.dna` file or not (as it has always been)...

We can decide to do that in a future PR if we want to... This PR opens the door for that.


govert:
This is awesome, thatnks @caioproiete !

It would be even nicer if there were a way to avoid adding the schema file to the project.

Is there another way to provide the schema file, e.g. by actually hosting it at that url location?
I presume it could also be installed in the schema cache of Visual Studio, for example in the (still supported) init.ps1 for of the package? Do you know if there is DTE object model support for this?

govert:
OK - the Ribbon .xsd files are here: https://www.microsoft.com/en-us/download/confirmation.aspx?id=4463 and https://www.microsoft.com/en-us/download/confirmation.aspx?id=1574 .
They're .exe files but unzip twice to get the .xsd files.
When installed in Visual Studio, it works fine with the nested customUI tag inside a .dna file.

I'm was hoping to edit the <Visual Studio>\Xml\Schemas\catalog.xml file to create an association between .dna files and the schema, so that it lights up even without the explicit xmlns. But that didn't seem to work the way I expect.

caioproiete:
@govert You're welcome.

>
> It would be even nicer if there were a way to avoid adding the schema file to the project.
>

Yes, I'd prefer not having to add the `.xsd` to the project as well. Ideally Visual Studio would be smart enough to see XSD schemas inside a NuGet package and just add it to its in-memory catalog, without having to add it to the project, but that's not the case.

There are [different ways to tell Visual Studio about which XSD file to use for validation and IntelliSense](https://docs.microsoft.com/en-us/visualstudio/xml-tools/xml-document-validation) but for NuGet packages the recommended way seems to be with `content` (`packages.config`) and `contentFiles` (`PackageRefrence`). For `PackageReference` I believe they support a way to have the file become part of the project (so that Visual Studio knows about it) but be hidden from the user in the project, which would work well for this scenario.

Adding the XSD to [Visual Studio's Schema Cache](https://docs.microsoft.com/en-us/visualstudio/xml-tools/schema-cache) manually is an option as you know, but I wouldn't think it's a good idea to do that from a NuGet package (_you'd need to be running VS as Administrator for a start, which not everyone can_) - nor would I recommend that as the Schema Cache is global to all projects, so a NuGet package changing settings that doesn't belong to the project they're being installed in, would be wrong.

I think having a new file added to the project to get IntelliSense and validation is a very small price to pay, and having it as part of `ExcelDna.AddIn` makes the experience better for new users that are just getting started with Excel-DNA.

Another option would be to create a separate NuGet package just with the XSD, and leave it to the user to install the additional NuGet package or not if he/she wants IntelliSense.

caioproiete:
>
> OK - the Ribbon .xsd files are here
>

Yep, I tested them as well together with the `.dna` schema, and it works just fine inside Excel DNA's `<CustomUI>` tag, as well as a separate XML (e.g. `Ribbon.xml`).

>
> I'm was hoping to edit the `\Xml\Schemas\catalog.xml file` to create an association
>

I tested this yesterday night and it worked for me. In addition to adding an entry to the catalog, did you copy the XSD file to the Schema Cache? (`%ProgramFiles(x86)%\Microsoft Visual Studio\[**XX**]\[Edition]\Xml\Schemas\` for VS2017 or `%ProgramFiles(x86)%\Microsoft Visual Studio [**XX**]\Xml\Schemas\` for \<= VS2015)

govert:
I want to think more about how to distribute the .xsd, and I don't think we should put it into the NuGet package for now. It's quite easy for a user to install 'manually', and maybe it should really go into a .vsix, and your idea of a separate NuGet package might be OK too. And it would be nice to add the Office .xsd at the same time too, but we need to decide how bad a license violation that would be.

I suggest this for the current version 0.35:
* We add the .xsd file to the Distribution folder so that it is easy to download from GitHub (or anywhere else you think is easy, or multiple places).
* We add the xmlns attribute to the standard .dna template file.
* Maybe we add a comment in the standard .dna template file that describes how to download and install the .xsd in Visual Studio. So it a user starts to edit the .dna file they will see the right instructions there.

We think about the other options for distributing or installing the .xsd later.

What do you think?

govert:
>I tested this yesterday night and it worked for me. 

Do you mean that you were able to open a .dna file that has no xmlns and it used the schema?
What was your catalog.xml entry?

govert:
For reference on how to install .xsd with a .vsix: https://docs.microsoft.com/en-us/visualstudio/extensibility/set-install-root

caioproiete:
>
> I want to think more about how to distribute the .xsd
>

Take your time. None of this needs to make it to v0.35 of course, but do you have any concern right now, though?

I mean, you know... It's just a text file, the user can open, read what's in there, etc. When uninstalling the NuGet package, it gets removed, etc. It doesn't cause any harm to the project or anything, if the user deletes the file after installing the NuGet package, everything works the same (just no IntelliSense, of course). A person developing an Excel-DNA add-in for the first time, gets a nice IntelliSense experience out-of-the-box from `File->New Project`... In my mind it really is a no-brainer, but maybe you're seeing something that I'm not (?).

If, for some reason we change our mind later, we can always remove it from the NuGet package in a future version. I'm trying to think of things that could go wrong with including an extra content file in the project, that doesn't get compiled or anything, and I don't see any.

>
> It's quite easy for a user to install 'manually',
>

Hmm, I don't think it's very easy for a junior developer to manage this, nor do I want my team to waste time with this whenever they create a new add-in, or get a new VM or whenever there's a new version of the `.xsd` because new properties got added to the `.dna` file, etc. Also not all developers have admin rights on their machine, and those devs cannot write to the Schema Cache, so they do need to install it to each project individually and that's error-prone, boring, and a waste of people's time.

It's 2018... Anything that is not a simple `install-package Bla` that gets me I'm up and running immediately, is bad IMO.

>
> maybe it should really go into a .vsix
>

VSIX would be a valid option, but it can easily get out of sync. We may require a newer schema version in the `.dna` template whilst the user didn't upgrade the VSIX on their machine yet, or they might have upgraded the VSIX but are still running and old version of the nuget package... Plus Visual Studio is already too slow, and I personally would refuse to install yet another VS extension just to get IntelliSense on a file that I don't change very often. In fact, if I could not use Visual Studio at all today, I would - hence my suggestion on the roadmap for VS Code, but I digress...


### The moment you need IntelliSense the most in `.dna` files, is after you just installed the `ExcelDna.AddIn` package in your project
It's when you're still getting your add-in set up and adding your dependencies, etc... After the initial project set up is done, changing the `.dna` file becomes a rare occasion over time. **That's why I think it should come bundled with `ExcelDna.AddIn`**, and "_just work (tm)_" from a clean install.

Having the XSD in a separate NuGet package **that is added as a dependency of `ExcelDna.AddIn`** so that it gets installed together, would yield the same result/UX so would work for me, but I don't see the appeal of that when the only place you're supposed to have `.dna` files, is on projects where `ExcelDna.AddIn` is also installed. _Unless we want to support multiple versions of the XML schema in the same file - but... really(?)_.

Having the XSD in a separate NuGet package that is **not** a dependency of `ExcelDna.AddIn` (and doesn't not install automatically) would be a compromise and we can go this route for a quicker time-to-market, but it's a missed opportunity to add a lot of value for a fresh install...

---

N.B.: Regardless of how we decide to distribute the XSD, we'll need to merge the changes to [`DnaLibrarySerialization.cs`](https://github.com/Excel-DNA/ExcelDna/pull/194/files#diff-5c602868defc97fa55f44298f6467b76) because it's currently hard-coded to expect an empty XML namespace and, because of that, adding any namespace to the `.dna` file breaks `ExcelDnaPack` and the build fails, so **this is crucial even for the manual XSD install approach**.

caioproiete:
> Do you mean that you were able to open a .dna file that has no xmlns and it used the schema?
> What was your catalog.xml entry?

Yep, these were the steps:

1. Copy `ExcelDna_dna_2018_05.xsd` to the Visual Studio's schema cache
  In my case it was `C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Xml\Schemas\`
2. Add `Association` entry to schema cache `catalog.xml`
  `<Association extension="dna" schema="%InstallRoot%/xml/schemas/ExcelDna_dna_2018_05.xsd" defaultNamespace="http://schemas.excel-dna.net/addin/2018/05/dna.xsd" enableValidation="false" />`
3. Create a new class library, and installed ExcelDna.AddIn v0.34.6
  image
4. Crack open a beer 🍺 

---

[Here is my full `catalog.xml` attached](https://github.com/Excel-DNA/ExcelDna/files/2019787/catalog.zip).

govert:
Ah - thanks. I was fiddling with directory from the wrong version. It also works to add a separate .xml file with only the one entry.

govert:
Should the namespace name itself perhaps not have the .xsd part at the end?

Is there a way to hide or indicate that Reference.AssemblyPath is deprecated? It's not too serious, just a small confusing nit.

caioproiete:
> 
> Should the namespace name itself perhaps not have the .xsd part at the end?
> 

Hmm, good catch! 👍 

It didn't seem to make any difference in practice (i.e. everything works either way) but to make it consistent with everything else, my steps above should change to:

1. Copy `ExcelDna_dna_2018_05.xsd` to the Visual Studio's schema cache **and rename it to `dna.xsd`**

2. Add `Schema` entry to schema cache `catalog.xml`
  `<Schema href="%InstallRoot%/xml/schemas/dna.xsd" targetNamespace="http://schemas.excel-dna.net/addin/2018/05" />`

3. Add `Association` entry to schema cache `catalog.xml`
  `<Association extension="dna" schema="%InstallRoot%/xml/schemas/dna.xsd" defaultNamespace="http://schemas.excel-dna.net/addin/2018/05" enableValidation="false" />`

image

---

[Revised `catalog.xml` attached](https://github.com/Excel-DNA/ExcelDna/files/2019821/catalog.zip).

govert:
No - I mean inside the .xsd file and in the .dna files.

caioproiete:
>
> No - I mean inside the .xsd file and in the .dna files.
> 

I don't think so...
https://github.com/Excel-DNA/ExcelDna/blob/master/Package/Excel-DNA/Excel-DNA.nuspec#L2

govert:
I see. It seems very unusual, if I compare with all the namespaces in the catalog.xml file, for example.

caioproiete:
> 
> I see. It seems very unusual, if I compare with all the namespaces in the catalog.xml file, for example.
> 

As long as it's something unique per each file type, it's all valid:
* ✅`http://schemas.excel-dna.net/addin/2018/05/dna.xsd`
* ✅`http://schemas.excel-dna.net/addin/2018/05/dna`
* ❌`http://schemas.excel-dna.net/addin/2018/05` 

I may be hard to believe, but in this particular case, I don't care either way 😃

govert:
OK - I understand the contents of the .xsd essentially specify the namespace identifier string, and the filename is incidental.

govert:
In the past the serializer class was generated by sgen (manually, when the schema changed). That was nice, because editing it will be error prone. I can leave the serializer class as is if I use a custom XmlTextReader, and that allows me to read with and without the namespace, but then the customUI breaks. I'll fiddle a bit more, but I don't have anything that's better than the manual edit you've made yet.

caioproiete:
>
> Is there a way to hide or indicate that Reference.AssemblyPath is deprecated? It's not too serious, just a small confusing nit.
>

AFAIK there's no way to do that in XSD 1.0, and Visual Studio doesn't support XSD 1.1 yet.

What we can do now is to just remove `AssemblyPath` from the XSD, so that it doesn't show in the IntelliSense, and the user will see an error message when opening the `.dna` file saying that the attribute is invalid, if it's being used.

govert:
OK, here's a first pass on how I'd like to proceed:

### Schema and core library changes

* The namespace we use is 'http://schemas.excel-dna.net/addin/2018/05/dnalibrary', without the .xsd extension in the namespace string, and matching the root element.
* We call the .xsd file 'ExcelDna.DnaLibrary.xsd'. We'll deal with versioning later.
* We remove the `Reference.AssemblyPath` from the xsd. Otherwise the schema looks great!
* We add the namespace to the `XmlRoot` attribute on the `DnaLibrary` class, so that using `DnaLibrary.Save` will add the namespace (only using the package .dna file currently). We want to be tolerant in what we read, and write with the xmlns included.
* We implement the manual edits to `DnaLibrarySerialization.cs`, with extra comments to explain what happened and what is edited from the sgen output. Our intention going forward will to either manually update, or carefully merge from sgen output. We should include the old command line in the comments there: "sgen /a:bin\Debug\exceldna.integration.dll /t:ExcelDna.Integration.DnaLibrary  /k /f /out:." 
* We remove the `CreateDnaLibrarySerializer.bat` file, so I won't press the button by mistake in a few years time.

### Schema distribution

* We create a new directory `\Distribution\XmlSchemas` under Distribution in the Git repo with five files:
  * `ExcelDna.DnaLibrary.xsd`
  * `customUI.xsd`
  * `customui14.xsd`
  * `ExcelDna.XmlSchemas.xml`
  * `Readme.md`

* `ExcelDna.XmlSchemas.xml` and the three schema files are intended to work well when copied into the VS \Xml\Schemas directory, including allowing the .dna schema to light up for .dna files that don't have an explicit schema, as you showed me.

* `Readme.md` explains what's in the directory and installation instructions. We'll point to the `\Distribution\XmlSchemas` folder from elsewhere as guidance, and expect `Readme.md` to light up in GitHub.

### NuGet packages

* We update the .dna file template in the `ExcelDna.AddIn` package to include the xmlns and with a comment added that links to the `\Distribution\XmlSchemas` folder on GitHub.

* The .xsd doesn't go in the ExcelDna.AddIn package for now, and we don't add a dependency that pulls it in automatically.

* We create a NuGet package called `ExcelDna.XmlSchemas` that includes the three .xsd files as content. It is versioned independently from the other ExcelDna packages (based on the contents) and has no dependencies either way. This will work for developers who don't have admin rights.

* Later I might try to create a small .vsix called ExcelDna.XmlSchemas or 'Excel-DNA XML Schemas' to install the schemas for developers with admin rights.

caioproiete:
@govert Sounds like a plan.

* What is the content of `ExcelDna.XmlSchemas.xml`?
* What about more descriptive names for `customUI.xsd` and `customui14.xsd`? e.g.:
  * `ExcelDna.Office2007.CustomUI.xsd`
  * `ExcelDna.Office2010.CustomUI.xsd`
  That would make them be all grouped together in the project tree at least, with same prefix

govert:
OK, it looks like the catalog file cannot have a "." in to work. (VS is supposed to automatically merge catalog files in \Xml\Schemas that end with .xml, but it only works for my with a simple name.)

I suggest instead of calling it `ExcelDna.XmlSchemas.xml` we call it `ExcelDnaCatalog.xml` with this content:

```xml
<?xml version="1.0" encoding="utf-8"?>
<SchemaCatalog xmlns="http://schemas.microsoft.com/xsd/catalog">
  <Schema href="%InstallRoot%/xml/schemas/CustomUI.xsd"                       targetNamespace="http://schemas.microsoft.com/office/2006/01/customui"          />
  <Schema href="%InstallRoot%/xml/schemas/customui14.xsd"                     targetNamespace="http://schemas.microsoft.com/office/2009/07/customui"          />

  <Association extension="dna" schema="%InstallRoot%/xml/schemas/ExcelDna.DnaLibrary.xsd" defaultNamespace="http://schemas.excel-dna.net/addin/2018/05/dnalibrary" enableValidation="false" />

</SchemaCatalog>
```

With that schema file I need not edit the real catalog.xml at all, and can just copy the .xml file and .xsd files into the Xml\Schemas directory.

I suggest we keep the exact names for the office .xsd files, even including the inconsistent capitalization, so that it is clear we are shipping the 'real' files and not our own version, and it they already exist we won't put duplicates. It's ugly, but I think it's then cleared to see what we're doing. I'm not sure what the exact license is for those files, but this feels the least offensive to me.

govert:
I don't know if we need the two `<Schema>` entries if we put the .xsd files into that directory, or whether we need an extra `<Schema>` entry for the `ExcelDna.DnaLibrary.xsd` to allow us to put it somewhere else.

Possibly we only need the `<Association>` entry?

govert:
OK - I edited the `ExcelDnaCatalog.xml` - we only need the association. Everything lights up if the four files are copied into `\Xml\Schemas`, including for .dna files with no schema.

govert:
> One thing: On the Association we need to set enableValidation="true" otherwise (as the name suggests) Visual Studio doesn't perform the schema validation (only IntelliSense suggestions/autocomplete).

Would failed validation of the .dna file block the build?

caioproiete:
>
> Would failed validation of the .dna file block the build?
>

No, XSD validation is not performed during the build, so no errors or warnings... The build always succeed regardless of the contents of the .dna file

---

ps: **If** you have the file open on Visual Studio, you see warnings, which is helpful:

image

govert:
I've created the .vsix extension (Excel-DNA XML Schemas) in the https://github.com/Excel-DNA/VSExcel repository.
You download and try the .vsix from the releases tab: https://github.com/Excel-DNA/VSExcel/releases
I'll publish as v1.0 to the marketplace once we're sure the schema is final.

caioproiete:
I think this is it for this PR, if I didn't miss anything.

### Schema and core library changes
* [x] _Use namespace `http://schemas.excel-dna.net/addin/2018/05/dnalibrary`_
* [x] _Call the .xsd file 'ExcelDna.DnaLibrary.xsd'_
* [x] _Remove the Reference.AssemblyPath from the xsd_
* [x] _Add the namespace to the XmlRoot attribute on the DnaLibrary class_
* [x] _Implement the manual edits to DnaLibrarySerialization.cs_
* [x] _Remove CreateDnaLibrarySerializer.bat file_

### Schema distribution
* [x] _Create a new directory `\Distribution\XmlSchemas`_
* [x] _Create `\Distribution\XmlSchemasExcelDna.DnaLibrary.xsd`_
* [x] _Create `\Distribution\XmlSchemascustomUI.xsd`_
* [x] _Create `\Distribution\XmlSchemascustomui14.xsd`_
* [x] _Create `\Distribution\XmlSchemasExcelDnaCatalog.xml`_
* [x] _Create `\Distribution\XmlSchemasReadme.md`_

### NuGet packages
* [x] _Update the .dna file template in the ExcelDna.AddIn package_
* [x] _Remove .xsd from ExcelDna.AddIn package for now_
* [x] _Create a NuGet package called `ExcelDna.XmlSchemas`_
* [x] _Final XSD for .vsix via https://github.com/Excel-DNA/VSExcel/pull/3_

ps: I think it makes sense to add the the XML namespace to all the samples in `/Distribution/Samples/`, but will leave it for a separate PR as this one is already quite big and don't to make review harder than already is.

govert:
@caioproiete  Wow - you are a machine!
I'll still have a look through, but I think that's everything.
Thank you!

I'm not sure the little .dna files in Distribution\Sample should get the namespace - or in general that _all_ .dna files should have the namespace in them. We'll see how it works. But we should discuss this more before you make the wholesale changes Those Distribution\Samples might move to the Excel-DNA\Samples repository over time anyway.

Distribution/XmlSchemas/Readme.md
```diff
@@ -0,0 +1,90 @@
...
+    * [`ExcelDna.DnaLibrary.xsd`][dna-xsd]
+    * [`customUI.xsd`][cui-2007-xsd]
+    * [`customui14.xsd`][cui-2010-xsd]
+    * [`ExcelDnaCatalog.xml`][dna-catalog-xml]
+
+2. Copy these files to your the Visual Studio Global Schema Cache. This will be `%ProgramFiles(x86)%\Microsoft Visual Studio [xx]\Xml\Schemas\` for Visual Studio 2015 and earlier or `%ProgramFiles(x86)%\Microsoft Visual Studio\[xx]\[edition]\Xml\Schemas\` for Visual Studio 2017 and later. `[xx]` will be the version such as `14.0` or `2017` and `[edition]` will be the edition such as `Community`, `Professional`, or `Enterprise`.
+
+    For more detailed information about [Visual Studio's Schema Cache][vs-schema-cache], see the documentation on Microsoft's website: [https://docs.microsoft.com/en-us/visualstudio/xml-tools/schema-cache][vs-schema-cache]
+
+3. Restart any open instances of Visual Studio. IntelliSense and validation should now work for all `.dna` files you open, automatically, even if they don't have the XML namespace for `.dna` files (`http://schemas.excel-dna.net/addin/2018/05/dnalibrary`) declared (although it a good practice to always include the XML namespace on your files).
```
govert: This is not needed.

Source/ExcelDna.sln
```diff
@@ -25,6 +25,16 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ExcelDna.IntegrationTests",
...
 	EndProjectSection
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Package", "Package", "{2A988F4A-6C24-49EB-950D-0984E2CAEE20}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "ExcelDna.AddIn", "ExcelDna.AddIn", "{42A3C1DE-5D65-4BC6-BA90-3FDEE42A77C9}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "content", "content", "{D0B02D05-F536-4C32-9DC0-BA2B5EEAEAB7}"
+	ProjectSection(SolutionItems) = preProject
+		..\Package\ExcelDna.AddIn\content\ExcelDna-Template.dna = ..\Package\ExcelDna.AddIn\content\ExcelDna-Template.dna
+		..\Package\ExcelDna.AddIn\content\ExcelDna_dna_2018_05.xsd = ..\Package\ExcelDna.AddIn\content\ExcelDna_dna_2018_05.xsd
```
govert: Not sure what if this file still exists.

Distribution/XmlSchemas/Readme.md
caioproiete: Hmm. Are you sure? It didn't work for me on FSharp projects until I restarted VS. I'll reword it as to "try it, and if doesn't work, restart" kind of thing

Source/ExcelDna.sln
caioproiete: Yep, good catch, this line needs to go.

govert:
@caioproiete  Given all we've said here, what do you see as the preferred way to get .dna IntelliSense:

* Install the ExcelDna.XmlSchemas NuGet package into your solution. (Messy but needs no admin rights and anyone else opening your project will get IntelliSense automatically.)

* Install the ExcelDna.XmlSchemas .vsix extension.  (Easy and works for existing .dna files, and all future projects. Needs admin rights. Needs a VS restart. Doesn't work for other users opening project.)

* Manually copy the four files from `Distribution\XmlSchemas` into the `Xml\Schemas` folder. (Somewhat tricky to install, but doesn't involve having an extra extension. Needs admin rights. Doesn't need a VS restart. Doesn't work for other users opening project.)

Anything about the pros and cons I'm missing?
What will you be doing internally?

caioproiete:
@govert

**TL;DR;**
I think you got all the main pros and cons of each approach. I see the NuGet package as the best approach of them all - and I'll go with that internally.

I also think the NuGet package story will get better once we start supporting `PackageReference` because we will be able to hide the files from the project tree, and it's not going to be messy anymore.

---

**Long version**:

I think you got all the main pros and cons of each approach. There's also the option to manually do what the NuGet package does, and add the files to the project ourselves after downloading from the `Distribution` folder on GitHub - same pros/cons as the NuGet approach except that without the benefits of the NuGet tool itself.

I'd also say that running `install-package ExcelDna.XmlSchemas` takes 5 seconds of a single developer vs. however many minutes it takes to read the Readme documentation, download the files, find the Schema Cache folder, copy the files, etc. multiplied by the number of developers in the team multiplied by how many machines each of them have - You can tell this is not my favorite :grinning: 

As it stands, in all 4 approaches things can get out of sync. If we change anything in the `DnaLibrary` class that requires updates to the XSD, users who upgrade to the newer version of `ExcelDna.AddIn` may get errors in the `.dna` files until they install the newer version of `ExcelDna.XmlSchemas`, or the newer version of the VSIX, or download the new files and copy them again to the Schema Cache or project, ...

In the future, if we ever decide to add `ExcelDna.XmlSchemas` as a dependency of `ExcelDna.AddIn`, then that would count as one more "pro" to the NuGet approach, because that would be the only case where things can never get out of sync (because we force the user to upgrade both packages together).

Does that make sense?

I'm a big believer that any developer should be able to clone my repository, open the VS solution, and everything "just works", getting the exact experience that I intended them to have when I pushed my code to source control. That means that _I'm careful to not have :tophat: "magic" :rabbit: stuff on my machine that others might not have or know about_.

If you take a look at several of my contributions to ExcelDna (#18, #16, #19 are some examples) you'll notice they all share this common theme of "scaling ExcelDna to work for development teams and helping developers hit the ground running when then clone an ExcelDna project from source control", so the NuGet package is the way to go for me and what I'd recommend to anyone...

govert:
I'm very happy to have an .xsd schema - thank you again.

For now I think my own preferred and recommended approach will be the .vsix extension. I'll try to update the guidance in a way that keeps it balanced and shows the different opinions, before the final release.

I'd prefer not to have a large-scale change of the samples before we discuss more.

PR #199, opened 2018-05-25
Better logging infra for build tasks

### LINKED ISSUE #190: SetDebuggerOptions fails when MSBuild nodes are reused between different Visual Studio instances
I've just noticed this on **Visual Studio v15.7.1**.

### Repro steps:

* Close any Visual Studio instances you might have opened
* Open a new Visual Studio instance (VS1)
* Create a new Class Library project (MyLibrary1)
* Install ExcelDna v0.35.1-beta2 into **MyLibrary1**
  * `Install-Package ExcelDna.AddIn -Version 0.35.1-beta2`
* Open a new Visual Studio instance (VS2)
* Create a new Class Library project (MyLibrary2)
* Install ExcelDna v0.35.1-beta2 into **MyLibrary2**
  * `Install-Package ExcelDna.AddIn -Version 0.35.1-beta2`
* Make sure there are no MSBuild processes running on the machine
  * `taskkill /im msbuild.exe /f /t`
    image
* Rebuild the solution **MyLibrary2**
  * This will start one `MSBuild.exe` process to compile the **MyLibrary2** (MB1)
  * **Everything should work as expected. No errors or warnings** :shipit:
* Now go to **VS1** and rebuild the solution **MyLibrary1**
  * Because the the `MSBuild.exe` process that we previously used to build **MyLibrary2** (**MB1**) is now idle, **VS1** will use it to build **MyLibrary1**
  * Although build finishes successfully, **you should see a warning `Unable to set the debugger options within Visual Studio`** :warning:
    image

---

This is happening because the way we get a reference to Visual Studio in order to set the debugger options is by [getting the parent process of the MSBuild process that is running our task](https://github.com/Excel-DNA/ExcelDna/blob/master/Source/ExcelDna.AddIn.Tasks/Utils/DevToolsEnvironment.cs#L73) which obviously fails in this scenario given that the parent process of MSBuild is **VS2** and the build task cannot find the project **MyLibrary1** inside the **VS2**. Doh!

### **The workaround, until we fix this, is to [disable MSBuild node reuse by defining an environment variable](https://github.com/Microsoft/msbuild/wiki/MSBuild-Tips-&-Tricks#msbuildexe-nrfalse)**

image

caioproiete:
Add some long overdue shared logging infrastructure inside the build tasks, so that all tasks, utilities and friends can participate in the same logging pipeline of the build (which helps me debug and fix #190)

### DIFF
--- a/Source/ExcelDna.AddIn.Tasks/AbstractTask.cs
+++ b/Source/ExcelDna.AddIn.Tasks/AbstractTask.cs
@@ -1,50 +1,10 @@
-﻿using System;
-using Microsoft.Build.Framework;
+﻿using Microsoft.Build.Framework;
 
 namespace ExcelDna.AddIn.Tasks
 {
     public abstract class AbstractTask : ITask
     {
-        private readonly string _targetName;
-
-        protected AbstractTask(string targetName)
-        {
-            if (string.IsNullOrWhiteSpace(targetName))
-            {
-                throw new ArgumentException("Value cannot be null or whitespace.", nameof(targetName));
-            }
-
-            _targetName = targetName;
-        }
-
         public abstract bool Execute();
-
-        protected void LogDebugMessage(string message)
-        {
-            #if DEBUG
-                LogMessage(message, MessageImportance.High);
-            #else
-                LogMessage(message, MessageImportance.Low);
-            #endif
-        }
-
-        protected void LogMessage(string message, MessageImportance importance = MessageImportance.High)
-        {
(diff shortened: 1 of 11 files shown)


govert:
@caioproiete Can you explain why it is preferable to log against this "DNA" + variable hashcode codes, rather than starting a list of our own codes, or even a fixed code: "DNA001" or whatever? Is there some convention or recommendation about this that you can point me to?

PR #204, opened 2018-06-06
Add option for including pdb files in packed addin

### LINKED ISSUE #111: Pack and load .pdb symbol files
We should add an option to the packing to include matching .pdb files if they are present, and unpack and load these in the AssemblyResolve.

See these discussions:
* https://groups.google.com/d/topic/exceldna/Wh6psWVJRMo/discussion
* https://groups.google.com/d/topic/exceldna/vZu0-O8goqM/discussion
* https://groups.google.com/d/topic/exceldna/fjUHTYiSREE/discussion

lanfeust69:
Main reason is to have nicer stack traces if the addin logs an exception.

Resolves #111 

Note that the sgen command I used to create `DnaLibrarySerialization.cs` produced a file with some differences regarding handling of the new namespace. As a result, significantly less places needed to be manually edited so that *.dna files both with and without a `xmlns` attribute could be parsed.

### DIFF
--- a/Package/ExcelDna.AddIn/content/ExcelDna-Template.dna
+++ b/Package/ExcelDna.AddIn/content/ExcelDna-Template.dna
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <DnaLibrary Name="%ProjectName% Add-In" RuntimeVersion="v4.0" xmlns="http://schemas.excel-dna.net/addin/2018/05/dnalibrary">
-  <ExternalLibrary Path="%OutputFileName%" ExplicitExports="false" LoadFromBytes="true" Pack="true" />
+  <ExternalLibrary Path="%OutputFileName%" ExplicitExports="false" LoadFromBytes="true" Pack="true" IncludePdb="true" />
 
   <!--
        The RuntimeVersion attribute above allows two settings:
--- a/Source/ExcelDnaPack/PackProgram.cs
+++ b/Source/ExcelDnaPack/PackProgram.cs
@@ -1,4 +1,4 @@
-﻿//  Copyright (c) Govert van Drimmelen. All rights reserved.
+//  Copyright (c) Govert van Drimmelen. All rights reserved.
 //  Excel-DNA is licensed under the zlib license. See LICENSE.txt for details.
 
 using System;
@@ -197,7 +197,8 @@ Other assemblies are packed if marked with Pack=""true"" in the .dna file.
 			string packedName = null;
 			if (integrationPath != null)
 			{
-				packedName = ru.AddAssembly(integrationPath, compress, multithreading);
+				// pdb is not in the nuget package, so chances are we want it if it is present (ExcelDna dev)// pdb is not in the nuget package, so chances are we want it if it is present (ExcelDna dev)	                // pdb is not in the nuget package, so chances are we want it if it is present (ExcelDna dev)// pdb is not in the nuget package, so chances are we want it if it is present (ExcelDna dev)
+				packedName = ru.AddAssembly(integrationPath, compress, multithreading, true);
 			}
 			if (packedName == null)
 			{
@@ -258,7 +259,7 @@ Other assemblies are packed if marked with Pack=""true"" in the .dna file.
 						}
 						else
 						{
-							string packedName = ru.AddAssembly(path, compress, multithreading);
+							string packedName = ru.AddAssembly(path, compress, multithreading, ext.IncludePdb);
 							if (packedName != null)
 							{
 								ext.Path = "packed:" + packedName;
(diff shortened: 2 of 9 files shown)


Package/ExcelDna.AddIn/content/ExcelDna-Template.dna
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <DnaLibrary Name="%ProjectName% Add-In" RuntimeVersion="v4.0" xmlns="http://schemas.excel-dna.net/addin/2018/05/dnalibrary">
-  <ExternalLibrary Path="%OutputFileName%" ExplicitExports="false" LoadFromBytes="true" Pack="true" />
+  <ExternalLibrary Path="%OutputFileName%" ExplicitExports="false" LoadFromBytes="true" Pack="true" IncludePdb="true" />
```
govert: I'm not sure this should be the default setting for a new project based on the NuGet package.

Source/ExcelDnaPack/PackProgram.cs
```diff
@@ -197,7 +197,8 @@ static void Pack(string[] args)
 			string packedName = null;
 			if (integrationPath != null)
 			{
-				packedName = ru.AddAssembly(integrationPath, compress, multithreading);
+				// pdb is not in the nuget package, so chances are we want it if it is present (ExcelDna dev)// pdb is not in the nuget package, so chances are we want it if it is present (ExcelDna dev)	                // pdb is not in the nuget package, so chances are we want it if it is present (ExcelDna dev)// pdb is not in the nuget package, so chances are we want it if it is present (ExcelDna dev)
+				packedName = ru.AddAssembly(integrationPath, compress, multithreading, true);
```
govert: I understand this to mean that we would not normally have the `ExcelDna.Integration.pdb` packed, but it a developer want to do so they could set it up so that `ExcelDna.Integration.pdb` is present, and then it would be packed.
I think I agree with this policy, but it should be documented somewhere.

Package/ExcelDna.AddIn/content/ExcelDna-Template.dna
lanfeust69: I dont't have a strong feeling here. It seems the only cost is a few more Kb, which is reasonable, and the upside is to show the feature (but this could also be achieved with the attribute present with its default value false).

Source/ExcelDnaPack/PackProgram.cs
lanfeust69: Yes, that's exactly the idea : in my case when debugging, I replace the reference to the Nuget package's assembly with a ProjectReference to my local clone of `ExcelDna.Integration.csproj`, in which case `ExcelDna.Integration.pdb` naturally comes along.
Not too sure where this should be documented, though...
And I'll still need to understand how the rebase duplicated the comment without me even noticing it !

lanfeust69:
I tested by following the `readme.txt` from `ExcelDna.AddIn.nupkg`, with this code :

```csharp
        [ExcelFunction(Description = "My first .NET function")]
        public static string HelloDna(string name)
        {
            LogSomeException();
            return "Hello " + name;
        }

        private static void LogSomeException()
        {
            try
            {
                throw new Exception("Exceptional !");
            }
            catch (Exception e)
            {
                File.WriteAllText(@"D:\exceldna.log", $"Exception caught : {e}");
            }
        }
```
Also make sure the original pdb is cleaned, else it will be used by default (at least when debugging in VisualStudio). Of course the real value is more when the stacktrace includes generated methods (lambda or async/await continuations), where the line numbers makes it much easier to understand.

BTW, any plan to have AppVeyor understand new C# syntax and provide meaningful feedback ?

govert:
@lanfeust69 Sorry - I'm still confused about why the `DnaLibrarySerialization` works with and without the `xmlns`. Can you explain how you generated it, and maybe help me understand why it worked for you when we had problems before?

lanfeust69:
Nothing fancy here : I just used the command mentioned in the comment, my sgen resolves to `C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.6.1 Tools\sgen.exe`, version 4.6.1055.0.

I saw that it was significantly different from the existing version, mainly with both an `id2_Item` containing the new namespace, and also an empty one in `id3_Item`.
So I first checked that everything seemed to work fine with the unmodified generated file, when the `xmlns` is present in the `.dna` file, then that as expected there was an error when the `xmlns` is missing. I then commented out the (only) 8 places where the namespace is checked against `id2_Item` (and failure of the test would cause an exception or a node to be skipped). As hoped, this allowed the unit tests to succeed, while of course still working with the `xmlns` present.

govert:
@lanfeust69 Thank you again for contributing this!
I've changed the default for IncludePdb in the template to "false", and pushed all of this as an updated set of NuGet packages too. Please let me know if you see that anything went wrong.

PR #231, opened 2019-03-08
Fix conversion wrapper for async functions

### LINKED ISSUE #230: Check bool parameters in native async functions
https://groups.google.com/forum/#!topic/exceldna/MwkfR6vsuw8

ittegrat:
If the async function throws an exception or an input parameter marshaler fails (e.g. DateTime or bool), the ExcelAsyncHandle is never informed that the function call is finished, keeping Excel calculating.
Fix #230


govert:
Wonderful thank you - this looks like the right plan.
I want to just test a little bit.

Issue #234, opened 2019-03-22
Async UDF with IsExceptionSafe = true flag


mvorisek:
Async UDF with `IsExceptionSafe = true` flag like:
```
        [ExcelFunction(IsExceptionSafe = true)]
        public static void exampleAsyncUdf(object value, ExcelAsyncHandle asyncHandle)
        {
            asyncHandle.SetResult((object)"res");
        }
```
will throw an Exception in the registration stage when the `ConvertToXlMethodInfos` method is called:
```
at System.Delegate.CreateDelegate(Type type, MethodInfo method, Boolean throwOnBindFailure)
at System.Delegate.CreateDelegate(Type type, MethodInfo method)
at ExcelDna.Loader.XlMethodInfo.CreateDelegateAndFunctionPointer(Type wrapperType)
at ExcelDna.Loader.XlMethodInfo.ConvertToXlMethodInfos(List`1 methods, List`1 targets, List`1 methodAttributes, List`1 argumentAttributes)
```

Is this a bug? Or is there any reason for it?

govert:
Thanks - I would call it a bug.
There are actually a few supported data types where the IsExceptionSafe=true causes trouble and registration fails. (I've often considered getting rid of IsExceptionSafe=true completely.)
In the long term I expect the whole marshalling story will get reworked and simplified.

PR #235, opened 2019-03-27
Fix async UDF with IsExceptionSafe = true

mvorisek:
Fix https://github.com/Excel-DNA/ExcelDna/issues/234 , type of the created delegate is then valid for given method directly and also the dynamically created wrappingmethod.


govert:
Cool @mvorisek  - thanks for fixing this :-)

PR #237, opened 2019-04-14
Fix sort for better perf

mvorisek:
The current sort optimization was not helpful if there were characters like "_" as the current sort was case invariant, but not lower case based.

### DIFF
--- a/Source/ExcelDna.Loader/XlRegistration.cs
+++ b/Source/ExcelDna.Loader/XlRegistration.cs
@@ -1,4 +1,4 @@
-﻿//  Copyright (c) Govert van Drimmelen. All rights reserved.
+//  Copyright (c) Govert van Drimmelen. All rights reserved.
 //  Excel-DNA is licensed under the zlib license. See LICENSE.txt for details.
 
 using System;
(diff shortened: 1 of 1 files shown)


mvorisek:
The answer is ASCII table.

Numbers are before `A` and `a`, so it does not matter if the string for comparison is in LC or UC.

But characters like `[` or the mentioned `_` (the only characters that can be used in UDF name outside alphanum) is between `A` and `a` and then it does matter if the case invariant sort uses LC or UC for the comparison.

It seems that Excel use an LC to compare, so we need to compare in LC too.

lanfeust69:
I think I understand what you mean : the sort must be done so that the resulting order matches what Excel will do. And it all boils down to what seems the only place where there can be a difference (given that the sort is case-insensitive) : whether '_' is "greater" or "smaller" than the letters.

Currently, the code sorts assuming '_' > 'a' (or 'A'). You're saying that Excel actually uses the other option. That's entirely possible, but you didn't mention how you tested this hypothesis. At least in the article referenced in the comment, I couldn't see an indication to that effect.

mvorisek:
@lanfeust69 Yes

govert:
Testing with 5000 functions "_", lower and uppers in the name, I get:
* No sorting: 22058 ms
* OrdinalIgnoreCase: 22385 ms
* Ordinal / ToLower: 21503 ms

govert:
Thanks @mvorisek

PR #250, opened 2019-07-05
Add strong name to Excel-DNA assemblies

### LINKED ISSUE #223: Add support for add-ins that use Strong-Name
As per https://groups.google.com/forum/#!topic/exceldna/hSCb4ig5rVI (and #76)

Currently, if we add a strong-name to a Class Library project that uses Excel-DNA, it doesn't work - i.e. The add-in doesn't get loaded because the copy of `ExcelDna.Integration.dll` that is embedded within the `.xll` doesn't have a strong-name, and we don't get a chance to add the strong-name ourselves, of course, because it's embedded in the `.xll`...

Probably the easiest approach is to just add a strong name to all of our assemblies, which includes `ExcelDna.Integration.dll`, so that future versions of Excel-DNA `.xll` add-ins ship with strong-named versions.

---

More info: [Strong naming .NET libraries](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/strong-naming#create-strong-named-net-libraries)

caioproiete:
This PR resolves #223 by signing the assemblies below with a strong-name:
- `ExcelDna.Loader.dll`
- `ExcelDna.Integration.dll`
- `ExcelDna.AddIn.Tasks.dll`
- `ExcelDnaPack.exe`

Having the assemblies above signed with a strong-name, allows developers to sign their own Excel-DNA add-ins with their strong-names.

Today:
```
Console.WriteLine(typeof(ExcelDnaUtil).Assembly.GetName().FullName);
// ExcelDna.Integration, Version=1.0.7056.37028, Culture=neutral, PublicKeyToken=null
```

After this PR is merged:
```
Console.WriteLine(typeof(ExcelDnaUtil).Assembly.GetName().FullName);
// ExcelDna.Integration, Version=1.1.0.0, Culture=neutral, PublicKeyToken=f225e9659857edb
```

---

* The strong-name key has been committed to the repository / publicly available as per [Microsoft's guidance](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/strong-naming#create-strong-named-net-libraries). See `Source\assets\exceldna.snk`.

* The assembly version has been fixed (instead of random generated) in order to reduce binding redirects, as per [Microsoft's guidance](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/strong-naming#create-strong-named-net-libraries). See `AssemblyInfo.cs` for the four projects mentioned above.

[image](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/strong-naming#create-strong-named-net-libraries)


govert:
Sorry @caioproiete  - I'm really concerned about possible problems with this. Every now and then I have to spend hours and hours sorting our the mess of "strict assembly loading" and binding redirects for other projects. I really see no benefit for Excel-DNA from strong naming at all. With string naming a user might install a version of ExcelDna.Integration into the GAC and then all add-ins will break.

Can't we just post some instructions for people to strong-name the ExcelDna.Integration themselves if they want to, with their own key?

caioproiete:
@govert, your concerns are understandable. Let's see if I can address them below:

> ... _mess of "strict assembly loading" and binding redirects_ ...

I get it and I feel this pain myself from time to time. This usually happens because of [diamond dependencies](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/dependencies#diamond-dependencies) when we have multiple versions of an assembly in the dependency tree.

This shouldn't be a problem for Excel-DNA itself, as we control all of our dependencies and can release them all at the same time as needed (forcing an upgrade on the dependencies via NuGet), so that part is covered.

I've also took the usual steps to mitigate assembly redirect pain:

1. [Freeze the `AssemblyVersion` from now on](https://github.com/Excel-DNA/ExcelDna/pull/250/files#diff-ab4654c1ba4ff6dae98e7173d352b001)
2. [Use `AssemblyInformationVersion` to track the real version number](https://github.com/Excel-DNA/ExcelDna/pull/250/files#diff-ab4654c1ba4ff6dae98e7173d352b001) (and, of course, versions are also reflected in the NuGet packages)
3. [Publish our strong name key and documentation to be used in any build](https://github.com/Excel-DNA/ExcelDna.StrongName)

Number **1** above means that the identity of assemblies (for the purposes of binding) will not change between versions, and will always be something like:

```
ExcelDna.Integration, Version=1.0.0.0, Culture=neutral, PublicKeyToken=f225e9659857edb
```

**That means no binding redirects required...**

Does that make sense?

---

> ... _benefit for Excel-DNA from strong naming_ ...

Strong-naming Excel-DNA enables:
* Developers that use Excel-DNA for building their own add-ins to sign their add-ins. Today this is not possible as described in #223
* Developers that extend Excel-DNA developing libraries (e.g. 1, 2, 3) to sign their libraries. Today this is not possible, as the libraries take a dependency on us

---

> ... _a user might install a version of ExcelDna.Integration into the GAC and then all add-ins will break_ ...

Well... OK 🤔 that's a correct statement, but a user would have to do that on purpose - and that's a mistake he/she would be doing.

How frequent do you think people would make this mistake? (and who uses the GAC in 2019 anyway? 😄) If we think this is a serious concern, we can easily add a simple check at the start, to see where the assembly was loaded from, at display an error message if it was from that GAC...

---

> _post some instructions for people to strong-name the ExcelDna.Integration_

Although this is certainly a possibility, it puts a big burden on _every developer_ using Excel-DNA and other libraries, and means they can no longer get Excel-DNA from nuget.

This is what a developer would have to do (at a minimum), in order to be able to strong-name their application today:
* Clone the Excel-DNA repository locally
* Add strong-name to `ExcelDna.Integration`
* Add strong-name to `ExcelDna.Loader`
* Build their own `ExcelDna.xll` and `ExcelDna64.xll` with the `ExcelDna.Integration` and `ExcelDna.Loader` embedded in them
* Somehow use these local-built files in their application, instead of a simple NuGet `install-package` 😞 

If the above sounds like a lot of work, now imagine if they use extensions of Excel-DNA, such as [`ExcelDna.Registration`](https://github.com/Excel-DNA/Registration), [`ExcelDna.IntelliSense`](https://github.com/Excel-DNA/IntelliSense), [`ExcelDnaDoc`](https://github.com/Excel-DNA/ExcelDnaDoc), etc... and any other 3rd-party extension for Excel-DNA such as [`ExcelDna.CellAddress`](https://www.nuget.org/packages/ExcelDna.CellAddress/), [`ExcelDna.Utilities`](https://www.nuget.org/packages/ExcelDna.Utilities/), [`Serilog.Sinks.ExcelDnaLogDisplay`](https://www.nuget.org/packages/Serilog.Sinks.ExcelDnaLogDisplay/), ...

They have to repeat the whole process above (clone, sign, etc.), to build locally all of these projects, because they all need a signed copy of `ExcelDna.Integration` (!)

We can save thousands of hours of developer's time and headaches, with ~20 lines of code...

govert:
@caioproiete Do you require strong named assemblies? Why?

To be clear, there is absolutely not relationship between .NET strong names and having a signed Excel .xll add-in. It would be nice to see some tool to ease signing the packed binary add-in (maybe as part of the build library) but this is an orthogonal issue to the one here.

No assembly that references `ExcelDna.Integration` is useful outside the context of an Excel-DNA add-in. So any such library can choose to not be a strong named library without it affecting other internal libraries. This means our libraries are different to something like `NLog` which is a general-purpose library that might well we used in contexts where signing is a sensible approach. 

To paraphrase you - who still uses assembly signing in 2019? And why? Fortunately it is not true that "every developer using Excel-DNA" needs a signed assembly - there have been perhaps two or three requests for this out of thousands of users over the last 15 years. It seems the only application for assembly signing is to follow cargo-cult corporate policies. I believe this is a legacy .NET feature that should be actively discouraged, and want to do my bit to help it die.

Excel-DNA interferes with the assembly loading in non-obvious ways, both to keep the `ExcelDna.Loader` private and implement the packing. So there might be unexpected side-effects of having signed assemblies in our particular context.

Signing also works completely different under .NET Framework and .NET Core. It part of the current .NET mess I want to avoid as long as possible.

It could be made easier for the user to sign the assemblies themselves - in particular it is not already necessary to rebuild any part of Excel-DNA in order to use a signed version of `ExcelDna.Integration`. Under circumstances we need to confirm, the packing will re-pack `ExcelDna.Integration` into the .xll. But this would take some guidance as to which assemblies to copy where and how to run the utilities.

But for now I just don't feel ready to pull in and support this change.

caioproiete:
@govert I don't require my team to add a strong name to our apps or add-ins, but I'm just lucky enough to be able to make those decisions myself at this point in time... Not every developer is in the same position.

As [Steve Jobs once said](https://www.youtube.com/watch?v=i5f8bqYYwps&t=1739), "_With the enterprise market, it's not so simple. The people that use the products don't decide for themselves, and the people that make those decisions sometimes are confused..._"

I don't disagree with the point that strong-name-signing assemblies of _applications_ is unnecessary and, as you say, follow cargo-cult corporate policies, but making it easier for devs to use Excel-DNA is more important to me, than being right. 

Strong naming the Excel-DNA assemblies reduces the barrier for using Excel-DNA for **some** users and that's a good thing... It's easy enough to get it done and tested in ~30 minutes, which is roughly what it took me to put this PR together.

It's [Microsoft's official guidance](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/strong-naming#create-strong-named-net-libraries) to **strong name publicly distributed libraries, to ensure the most people can use it**.

[image](https://youtu.be/hLFyycJVo0I?t=1584)

---

Adding a strong name only changes the identities of the assemblies... There is no other side-effects even with all the magic that happens with `ExcelDna.Loader`. Fun fact: historically, the identities of every assembly of Excel-DNA has changed in every release, because we're using `AssemblyVersion("1.0.*")` so the identity has been completely random all this time, without any issues...

DanZee:
> No assembly that references ExcelDna.Integration is useful outside the context of an Excel-DNA add-in. So any such library can choose to not be a strong named library without it affecting other internal libraries. 

The code that we use with ExcelDNA is part of a solution that builds a collection of shared libraries, a few executables and two Excel add-ins. The application needs to be signed, which means that all shared libraries need to be signed, which means our unit test assemblies need to signed (because of how InternalsVisibleToAttribute works), which means the code that references ExcelDNA needs to be signed, which means ExcelDNA needs to be signed.

Therefore, we unpack the ExcelDna nuget packages, sign the assemblies, repack them and then host the new packages on our internal nuget server. 

This is a minor inconvienence. However, I do not understand what the big deal is with just signing the assemblies, all other external libraries that we use (and that's quite a few) are strong named "out of the box". We have not experienced any technical issues related to signing.

> there have been perhaps two or three requests for this out of thousands of users over the last 15 year

There are 3 options for potential users with problems similar to ours:
1. Figure out how to sign unpack/sign/repack
2. Move on to some other product
3. Complain here
I would guess that option 2 is the most popular one.

govert:
OK guys - I'll think about it a bit more.

I'd appreciate any other feedback from users who would find this useful.

buybackoff:
> I'd appreciate any other feedback from users who would find this useful.

Having a couple of additional lines in `.csproj` files and a public signing key inside a repo is the best practice currently. I do not understand so frequent opposition against signing OSS .NET libs - it adds zero costs to those who do not care, but often a blocker for those who do care. I suffered many times and had to repack some libs. Not with ExcelDna so far, but I could easily imagine the scenario with `InternalsVisibleTo` and unit tests, this is a pretty valid setup that requires a strong name despite the fact that ExcelDna is the top dll in the final add-in.

govert:
OK - let's try this for a release or two. Despite my opposition and concerns about the assembly redirect mess, (and clearly "frequent opposition" by other too) it has been an issue raised by users from time to time. Let's try it and see if strong-naming results in fewer questions. If the redirects make a mess later, we can revert back to an unsigned release, or fork into a separate package with signed assemblies (though I know this is not recommended).

Some changes I would like to see:

1. We should only strong-name the assemblies that add-in libraries will reference, (not the 'internal' Excel-DNA assemblies like ExcelDnaPack.exe) i.e. only sign
   * ExcelDna.Integration.dll
   * ExcelDna.Registration(.<lang>).dll
   * ExcelDna.IntelliSense.dll

2. The fixed version should rather be v1.1.0.

caioproiete:
@govert: In interest of making progress on this one, I'm happy to update the PR to address items (1) and (2) above.

I'll keep the `IntegrationTests` project signed though, otherwise we will be blocked from writing tests to any internal members of `ExcelDna.Integration` due to the way [`InternalsVisibleTo`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute?view=netframework-4.8#examples) works. For instance, I'd love to write tests for [`AssemblyLoader`](https://github.com/Excel-DNA/ExcelDna/blob/master/Source/ExcelDna.Integration/AssemblyLoader.cs#L18) at some point, to give us more confidence in making changes to it.

---

ps: Out of curiosity, is there a technical reason not to sign `ExcelDnaPack.exe` and other assemblies? It's a [common](https://github.com/npgsql/efcore.pg/blob/dev/Directory.Build.props#L30-L33) [practice](https://github.com/search?q=org%3Aaspnet+Directory.Build.props+SignAssembly&type=Code) to sign all assemblies in a project with the same key (usually done via `Directory.Build.props` nowadays so new projects added in the future get automatically signed without any extra effort).

govert:
@caioproiete After think about it, my only concern is that ExcelDna.Loader.dll must not get signed. It's loaded in a context where this kind of change can be problematic, and signing it gives no advantage to an add-in developer.

ExcelDnaPack.exe probably makes no difference either way, and if you think signing IntegrationTests might help that's fine.

caioproiete:
@govert This should be good to go.

PR #263, opened 2019-10-25
Add `ExcelIntegration.UnhandledException` event

augustoproiete:
TL;DR; This PR adds an event called `UnhandledException` to `ExcelIntegration`, as a soft replacement to `ExcelIntegration.RegisterUnhandledExceptionHandler` (now marked as obsolete).

---

We can currently catch unhandled exceptions that are thrown from within Excel-DNA functions or macros by calling `ExcelIntegration.RegisterUnhandledExceptionHandler`:

```csharp
ExcelIntegration.RegisterUnhandledExceptionHandler(ExcelDnaUnhandledException);
```

This PR adds the ability to register handlers via an event, instead of being a method call, which is more natural and aligns with the way these handlers are registered in regular .NET apps... For example:

```
// Any app
AppDomain.CurrentDomain.UnhandledException += AppDomainUnhandledException;
TaskScheduler.UnobservedTaskException += TaskSchedulerUnobservedTaskException;

// WinForms apps
Application.ThreadException += ApplicationThreadUnhandledException;

// WPF apps
Dispatcher.CurrentDispatcher.UnhandledException += DispatchedUnhandledException;
```

---

### Usage:

```csharp
public class MyExcelAddIn : IExcelAddIn
{
    public void AutoOpen()
    {
        ExcelIntegration.UnhandledException += ExcelDnaUnhandledException;
        // ...
    }

    public void AutoClose()
    {
        // ...
    }

    private object ExcelDnaUnhandledException(object exceptionObject)
    {
        // ...
        return "Sorry!";
    }
}
```

---

This PR doesn't introduce any breaking changes. Ideas/improvements we could consider in the future (which would introduce breaking changes):

* Modify the delegate signature to have an `EventArgs` argument with a property that specify the source of the exception (Function, Macro, etc.)
* Modify the delegate signature to have an `EventArgs` argument with a property that holds the return value (and modify the delegate return type to `void`) for cases where multiple event handlers are registered and can access/modify values written by previous handlers
* Modify the delegate signature to have an `EventArgs` argument with a property that indicates if the exception was handled or not (and let it bubble up if not)

### DIFF
--- a/Source/ExcelDna.Integration/ExcelIntegration.cs
+++ b/Source/ExcelDna.Integration/ExcelIntegration.cs
@@ -33,6 +33,8 @@ namespace ExcelDna.Integration
 
         private static TryExcelImplDelegate tryExcelImpl;
 
+        private static readonly object _syncLock = new object();
+
         [UsedImplicitly] // called by reflection
         internal static void SetTryExcelImpl(TryExcelImplDelegate d)
         {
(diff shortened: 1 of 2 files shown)


govert:
Isn't it weird to have an event that returns something?
That's not what I would expect, and is important in this case.

caioproiete:
@govert Conceptually, yes, it's unusual... But having used Excel-DNA's current error handling, it doesn't shock me as the delegate has always been like that.

Want to _bite the bullet_ now and do it a little better (despite the breaking change)?

```csharp
[Serializable]
public delegate void ExcelDnaUnhandledExceptionEventHandler(object sender, ExcelDnaUnhandledExceptionEventArgs e);

[Serializable]
public class ExcelDnaUnhandledExceptionEventArgs : EventArgs
{
    public ExcelDnaUnhandledExceptionEventArgs(Exception exception)
    {
        Exception = exception ?? throw new ArgumentNullException(nameof(exception));
        ReturnValue = ExcelError.ExcelErrorValue;
    }

    public Exception Exception { get; }
    public object ReturnValue { get; set; }
}
```

We could go one level further and create separate delegates and events for unhandled exceptions in functions and macros, for example. Now or later...

govert:
I really don't think this is a change worth making, certainly not to obsolete the current method.
If we were doing it for the first time, it might well be a better design, but there's a higher bar for making a change now.
Functionally it doesn't really make a difference (adding multiple handlers from different parts of your add-ins seems like a very bad idea, counting against the event approach). It doesn't save any code.
The behaviour required (settings a value that is the result of the function) is weird for enough for an event that it doesn't really fit into the usual reporting-only UnhandledExceptionHandler events from other places.

I don't think it easier to discover what to do for a correct implementation and the existing method is at least mentioned in some posts and some examples.

Finally, if you're making lots of add-ins and have a string opinion about the event approach, it is trivial to add the event approach to a utility or helper class, and then use it that way in your add-ins.

Issue #265, opened 2019-10-28
Excel-DNA.Interop vs background threads


fandrei:
It seems that calls to COM interop interfaces or methods from background threads often finish without problems but result in Excel crashing in random ways some time later; I wonder if there is any way to make Excel-DNA.Interop throw an exception immediately when it's called from any thread besides the main thread?

garethhayter:
I always access the Application object via my own property, where I store the thread ID of the UI thread, and compare the running thread id with that, throwing an exception when they don’t match, as this means it’s being accessed from a non-UI thread.

 

This check can be wrapped in an “#if DEBUG” pre-processor directive so that it is not checked in release mode – you should be finding issues at dev time.

govert:
One problem is that you might make the `ExcelDnaUtil.Application` call to get the Application object on the main thread, then store it is some variable that you access and use from another thread (even implicitly by capturing a COM object in a delegate).

fandrei:
@govert agreed. And even worse, same problem applies to any other interface that may be obtained from the API and stored somewhere, explicitly or implicitly.

fandrei:
@govert by the way is it open source?

fandrei:
@caioproiete do you know whether ExcelDna.Interop is open source?

caioproiete:
@fandrei **No, `ExcelDna.Interop` is not open-source**. The assembly is the equivalent to the one auto-generated by Visual Studio when you add a COM reference to `Microsoft Excel (version) Object Library`.

![image](https://user-images.githubusercontent.com/177608/67920094-a17a2380-fb81-11e9-9cad-77b281f1989c.png)

govert:
The interop assemblies in `ExcelDna.Interop` match Excel 2010. The binaries are from the Primary Interop Assemblies redistributable from Microsoft, but can be re-created by running [TlbImp](https://docs.microsoft.com/en-us/dotnet/framework/interop/how-to-generate-primary-interop-assemblies-using-tlbimp-exe) on the Office .tlb type library definitions.

As far as I know, it's not quite possible to create the same interfaces using C# code that wraps the COM calls internally (and can thus do something like check the threads before forwarding the calls). Some magic in the IL or the way the PIA assembly is referenced means the real PIA is treated differently when referencing it, allowing some extra features.

[...]

PR #267, opened 2019-10-31
Expose `ManagedThreadId` of thread where the add-in was loaded

caioproiete:
Relates to #265. Allows developers to get access to the `ManagedThreadId` where the Excel-DNA Add-in was loaded from, so they can determine if it's safe to make COM calls or not.

### DIFF
--- a/Source/ExcelDna.Integration/Excel.cs
+++ b/Source/ExcelDna.Integration/Excel.cs
@@ -70,6 +70,11 @@ namespace ExcelDna.Integration
             }
         }
 
+        public static int MainManagedThreadId
+        {
+            get { return _mainManagedThreadId; }
+        }
+
         #region Get Window Handle
         // NOTE: Careful not to call ExcelVersion here - might recurse causing StackOverflow
 


govert:
An alternative (maybe better?) would be to expose the internal method `ExcelDnaUtil.IsMainThread`. 

You should note that being on the main thread is necessary but not sufficient evidence to mean it's safe to make reliable COM calls. The current check is called `ExcelDnaUtil.IsApplicationOK` which actually makes a call to the `Application` object.

govert:
@caioproiete I think my question is one of style - do you think `ExcelDnaUtil.ManagedThreadId` (which is always 1) is a better or a worse API to expose than `ExcelDnaUtil.IsMainThread` (which represents the way it would be used) or `ExcelDnaUtil.IsApplicationOK` which might be the intended question? 

Or do we not expose any of these internals, and just document the benefit you note of calling `ExcelDnaUtil.Application` when needed and not caching it some field?

caioproiete:
@govert Re: `ManagedThreadId` vs `IsMainThread`, I don't see it as a matter of "_style_" but more of "_good_" architecture vs "_not-so-good_" architecture...

I believe calling `ExcelDnaUtil.IsMainThread` (from an add-in) is **never** a good idea because it's a **_dynamic_ static property that can't be mocked**, _and_ the add-in author have no control over its implementation... Today it's safe to call it from anywhere, sure, knowing that (currently) it can return `1` or `0` (\*), but the add-in author has no way to guarantee it will continue to be safe to call from anywhere or that it will always have the same behavior in future versions.

On the other hand, `ExcelDnaUtil.ManagedThreadId` gives us a handy `int` that we can cache at the start of the add-in (i.e. `AutoOpen`), and never have to call it again... Much better (IMO).

As an example, nowadays I get `Thread.CurrentThread.ManagedThreadId` inside `AutoOpen` (which in `v1.1.0` I will conveniently replace with `ExcelDnaUtil.ManagedThreadId`) and register it with my [IoC container](https://autofac.org), so that services can get it later, without having to depend on `ExcelDnaUtil` anymore (as **this would mean my services (in other projects in the solution) would have to take a dependency on `ExcelDna.Integration`** which is a no-no).

---

(\*) Calling `ExcelDnaUtil.ManagedThreadId` outside of a context of an Excel add-in (e.g. from a unit test) will return the default value for `int`, of course, as it has not been initialized.

govert:
@caioproiete OK thanks for explaining how you think about this. You're wrong about "good" architecture, but because you're more stubborn then me, I'll let you have this one ;-)

We'll leave it with ManagedThreadId exposed if you'll find it useful.

caioproiete:
😃 hahaha jokes aside, I'll aiming to set aside some time in the evenings over the next week or so to see if I can bring [`ExcelDna.AddIn.Abstractions`](https://github.com/caioproiete/exceldna-addin-abstractions) back to life and with some examples, and it would great to have your input there.

Having everything as instance classes with implemented interfaces (or abstract classes with virtual members) makes things better, and with the added benefit of being able to extend the API with extension methods (by the add-in developer).

govert:
Not sure what "jokes" you mean . . .

An extra library (inside or outside the Excel-DNA organization) that provides this kind of alternate interface to the Excel-DNA core features might relieve some of the design tensions, and can evolve on a faster cadence.

PR #271, opened 2019-11-02
Check if Application is cached before `GetApplicationFromWindows`

caioproiete:
I'm might have missed something here (correct me if I did), but from reading the code and comments, my understanding is that there used to be a reason to assume that if `ExcelDnaUtil.Application` was being called from a non-UI thread, then it would be impossible for it to be cached already (from a previous call on the UI thread), because of `ThreadStatic` - but this is no longer the case.

This lets the code continue the execution flow (with a call to `IsApplicationOK`) if a previous call to `Application` succeeded, rather than trying to `GetApplicationFromWindows`, just because it's a non-UI thread.

### DIFF
--- a/Source/ExcelDna.Integration/Excel.cs
+++ b/Source/ExcelDna.Integration/Excel.cs
@@ -229,9 +229,12 @@ namespace ExcelDna.Integration
                             "Cross-thread operation not valid: Application accessed from a thread other than the thread it was created on.");
                     }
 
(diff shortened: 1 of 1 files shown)


govert:
@caioproiete  You are fiddling with a part of the code that has been extremely problematic historically and has taken up a significant part of the support and development time in the past. Really.

This may or may not be a safe change, but I con't have the cycles to think about it and see not significant benefit to this change, whether it be a safe change or not.

Issue #286, opened 2019-12-28
Construct ExcelReference from collection of other ExcelReferences


n1l:
It could be useful to extend ExcelReference c'tor with such parameter if we are talking about the same sheetId.

//code sample
ExcelReference collatedReference = new ExcelReference(
                        references.Select(r => new[] { r.RowFirst, r.RowLast, r.ColumnFirst, r.ColumnLast}).ToArray(), sheetId);

but it could be better to have saometing like this:
ExcelReference collatedReference = new ExcelReference(references, sheetId);

I've created a pr for this.

govert:
Closed by #288

PR #288, opened 2019-12-28
Extend ExcelReference constructor

n1l:
Issue with details - https://github.com/Excel-DNA/ExcelDna/issues/286

### DIFF
--- a/Source/ExcelDna.Integration/ExcelReference.cs
+++ b/Source/ExcelDna.Integration/ExcelReference.cs
@@ -122,12 +122,24 @@ namespace ExcelDna.Integration
             }
         }
 
-        public ExcelReference(IEnumerable<ExcelReference> references, IntPtr sheetId)
+        public ExcelReference(IEnumerable<ExcelReference> references)
         {
             int refCount = 0;
+            IntPtr sheetId = IntPtr.Zero;
             Dictionary<int, int[]> rectangles = new Dictionary<int, int[]>();
+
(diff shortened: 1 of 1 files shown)


govert:
Thanks for this.

* I think we should add a check and throw an exception if the SheetIds are not all the same.

* Where do you get the ExcelReferences that make this a useful override for you?

n1l:
We had a performance issue while changing cell format, so we create some code to cache references before format changing to improve performance.

n1l:
About exception, the code is basically copied from another c'tor with array in params, so in the code we use only rectangle values from references, and the sheetId is the same.

govert:
Since an ExcelReference refers to a range on a particular sheet, I think it would be a mistake to send in ExcelReferences that refer to different sheets (since the compound ExcelReference cannot support these). So I suggest we:
* Change the signature to only be 
`public ExcelRefernece(IEnumerable<ExcelReference> references)`
* Throw an `ArgumentException` if there are different SheetIds among the references, and use the unique `SheetId` for the `ExcelReference` we're constructing.

An alternative would be a static helper on the `ExcelReference` class like
`public static IEnumerable<ExcelReference> UnionPerSheet(IEnumerable<ExcelReference> references)`
that groups per `SheetId` and returns a compound `ExcelReference` for each `SheetId'. I'm not sure whether that's better, and in any case a utility library in the add-in can do this easily too.

n1l:
I see, I think you're right. I've updated c'tor.

PR #307, opened 2020-03-21
Add `XllPathInfo` property to `ExcelDnaUtil`

caioproiete:
This PR adds a property called `XllPathInfo` to `ExcelDnaUtil` of type `System.IO.FileInfo`. I'm thinking that most of the time, when add-in developers use the property `XllPath`, they are interested in getting the **folder** where the `.xll` is located, in order to access a file that is located on the same folder as the `.xll`.

`XllPathInfo` then becomes an alternative to `XllPath` that also includes a `DirectoryName` property for easy access of the add-in install folder, allowing this:

```csharp
var xllPath = ExcelDnaUtil.XllPath;
var xllFolder = Path.GetDirectoryName(xllPath);
var excelFilePath = Path.Combine(xllFolder, "File.xlsx");
```

To to become this:

```csharp
var excelFilePath = Path.Combine(ExcelDnaUtil.XllPathInfo.DirectoryName, "File.xlsx");
```

---

I'm open to ideas...

- Is `XllPathInfo` a good choice of name? Should we use a longer name like `XllPathFileInfo`?
- Should we go for `XllDirectoryInfo` (or diff name) instead, returning a `DirectoryInfo` instead of `FileInfo`, therefore exposing the installation folder specifically? Why is that better?

### DIFF
--- a/Source/ExcelDna.Integration/DnaLibrary.cs
+++ b/Source/ExcelDna.Integration/DnaLibrary.cs
@@ -50,6 +50,16 @@ namespace ExcelDna.Integration
             }
         }
 
+        private static FileInfo _xllPathPathInfo;
+        [XmlIgnore]
+        internal static FileInfo XllPathInfo
+        {
+            get
+            {
+                return _xllPathPathInfo;
+            }
+        }
+
         private string _Name;
         [XmlAttribute]
         public string Name
@@ -408,6 +418,7 @@ namespace ExcelDna.Integration
 
             // CAREFUL: Sequence here is fragile - this is the first place where we start logging
             _XllPath = xllPath;
+            _xllPathPathInfo = new FileInfo(xllPath);
             Logging.LogDisplay.CreateInstance();
             Logger.Initialization.Verbose("Enter DnaLibrary.InitializeRootLibrary");
             byte[] dnaBytes = ExcelIntegration.GetDnaFileBytes("__MAIN__");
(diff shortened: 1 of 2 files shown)


govert:
@caioproiete  Honestly, I think this is a pointless one-liner. Obviously you need to whole add-in path including the xll name, and that property matches what the C API.
If you'd really use this yourself, I'm happy to merge it in, though.

caioproiete:
@govert Why do you think it's pointless? [Searching for `ExcelDnaUtil.XllPath` on GitHub](https://github.com/search?o=desc&q=ExcelDnaUtil.XllPath&s=indexed&type=Code) to understand how people are using it tells me that most of the time people are using it only find out the folder where the add-in is located.

A few random examples from different projects: 

- [gjvdkamp/SimpleImageMsoViewer](https://github.com/gjvdkamp/SimpleImageMsoViewer/blob/1ce900751b8b851c8581436a29dab6a4bb17c6f5/SimpleImageMsoViewer/AddIn.cs#L28)
- [DanWBR/dwsim4](https://github.com/DanWBR/dwsim4/blob/ba2f7204a32702e3f3f6f0d9ee91cc2ca8e1c084/DWSIM/Assets/DWSIM.dna#L22)
- [m1sta/BlackBear](https://github.com/m1sta/BlackBear/blob/bd1fa6cb5224374d7f30f6516a234030feaf01e4/BlackBearAddin/CustomXML/UserSettings.cs#L55)
- [WCGPR0/parsers](https://github.com/WCGPR0/parsers/blob/c3af6faa04766a63ec5ca160e8a29fc13e76fd3f/NET/ConstantContactSendTrades/ConstantContactSendTrades/GlobalFunctions.cs#L163)
- [HongchenMeng/ExcelDnaSamples](https://github.com/HongchenMeng/ExcelDnaSamples/blob/32c05d72d5538c7dbecdf3ba4dcf218b398ecb82/AddInReloader/AddIn.cs#L19)

PR #309, opened 2020-04-13
Documentation Revival

rkapl123:
Dear Govert!

As discussed, the whole revived codeplexx documentation revived on Github, I've also weeded out lots of unreachable links (whether the content of the reachable links is still of value, I can't really estimate) and started building a Reference section (from the beginnings of the Reference page).

I also have some ideas concerning the reference documentation of key classes and types, but we can discuss that by mail, if you prefer.

-regards,
Roland


govert:
OK great - thank you. I'll try to clean up the front page a bit some time to get rid of the broken links and extra web-site stuff, to only leave the documentation pages.

Issue #193, opened 2018-05-18
Update ExcelDna.AddIn NuGet Package to support the new PackageReference format


augustoproiete:
Spin-off of #138 to discuss how we want to support the `PackageReference` format in the `ExcelDna.AddIn` NuGet package. This relates to #188

### Background

* Currently `ExcelDna.AddIn` relies on PowerShell scripts for [installing](https://github.com/Excel-DNA/ExcelDna/blob/master/Package/ExcelDna.AddIn/tools/install.ps1) and [uninstalling](https://github.com/Excel-DNA/ExcelDna/blob/master/Package/ExcelDna.AddIn/tools/uninstall.ps1) the package which basically do the following:

#### `install.ps1`
1. Rename an `_UNINSTALLED_$(ProjectName)` to `$(ProjectName)-AddIn.dna`, if it exists
_Which would have happened after the user uninstalled a previous version of ExcelDna_
2. Add a `$(ProjectName)-AddIn.dna` file, if step 1 didn't find an old file
2.1. Perform some string replacements inside the `.dna` file added to include `$(ProjectName)`
3. ~Add a reference to the `.targets` file~ **Update 2018-05-18**: This is now done automatically by a NuGet feature implemented by #167

#### `uninstall.ps1`
1. Rename an existing `$(ProjectName)-AddIn.dna` file to `_UNINSTALLED_$(ProjectName)`, if it exists

### Known issues

* [The `PackageReference` format does not support the `install.ps1` and `uninstall.ps1` PowerShell scripts](https://blog.nuget.org/20170316/NuGet-now-fully-integrated-into-MSBuild.html#will-existing-nuget-packages-work-if-i-am-using-packagereference-for-wpf-windows-forms-or-aspnet-projects)
* [The `PackageReference` format also does not support `content` files](https://blog.nuget.org/20160126/nuget-contentFiles-demystified.html#supported-project-types) the same way they do with the `packages.config` format and introduce another feature `contentFiles` that behaves in a different way where [files are meant to be immutable](https://docs.microsoft.com/en-us/nuget/reference/nuspec#including-content-files) and therefore [they are not shown in the project tree, and are not editable](https://github.com/NuGet/Home/issues/4954). The (terrible IMO) recommendation from the NuGet team is to "[_... copy them to the project manually if you need to edit the files ..._](https://github.com/NuGet/Home/issues/4954#issuecomment-337321237)" :disappointed:
[...]

PR #318, opened 2020-06-13
Improve ExcelDna.Integration `CopyLocal=false` story

### LINKED ISSUE #188: Update ExcelDna.Integration NuGet Package to support the new PackageReference format
Spin-off of #138 to discuss how we want to support the `PackageReference` format in the `ExcelDna.Integration` NuGet package.

### Background

* When a user installs the `ExcelDna.AddIn` NuGet package it comes with XLL templates that includes the `ExcelDna.Integration` assembly embedded in it
* The `ExcelDna.AddIn` NuGet package NuGet package also depends on the `ExcelDna.Integration` NuGet package, which brings a copy of the `ExcelDna.Integration` assembly and references it in the project - necessary for developing add-ins
* Currently `ExcelDna.Integration` relies on a [PowerShell script that sets the reference to `ExcelDna.Integration` to be `CopyLocal=false`](https://github.com/Excel-DNA/ExcelDna/blob/master/Package/ExcelDna.Integration/tools/install.ps1) to make sure this assembly doesn't get copied to the `$(TargetDir)` because we want to use the version that's embedded in the XLL template that comes with `ExcelDna.AddIn` - and [historically it caused problems to have `ExcelDna.Integration.dll` in the same folder](https://github.com/Excel-DNA/ExcelDna/blob/master/Source/ExcelDna.Integration/Project.cs#L227) - this may be something to revisit in a different issue in the future, but for the purposes of this issue let's keep the same behavior we have today which is **make sure we don't have a copy of `ExcelDna.Integration.dll` in the output folder of an Excel Add-In**

### Known issues

* [The `PackageReference` format does not support the `install.ps1` PowerShell script that runs during installation of the package](https://blog.nuget.org/20170316/NuGet-now-fully-integrated-into-MSBuild.html#will-existing-nuget-packages-work-if-i-am-using-packagereference-for-wpf-windows-forms-or-aspnet-projects)
* Developers often need to reference the `ExcelDna.Integration` assembly in other projects, such as Unit Test projects, and **in those cases they actually need a copy of `ExcelDna.Integration.dll` in the output folder of their unit test project**

### Proposed solution

I believe the `CopyLocal=false` behavior should be a concern of the `ExcelDna.AddIn` package instead, and not a concern of `ExcelDna.Integration`.

As part of adding support for `PackageReference` format, we should remove the `install.ps1` from the `ExcelDna.Integration` and projects that add this NuGet package to their projects will get a regular reference to `ExcelDna.Integration` as expected (which fixes the issue for unit test projects, etc.).

We'll let the `ExcelDna.AddIn` NuGet package handle enforce the  `CopyLocal=false` behavior itself.

Thoughts?

/cc @govert @screig

augustoproiete:
Add a build target in the `ExcelDna.AddIn` build process that ensures `ExcelDna.Integration.dll` is **not** copied to the output folder, effectively removing the need for setting `CopyLocal=false` on Excel-DNA Add-in projects.

This also removes the need for the `install.ps1` in the `ExcelDna.Integration` NuGet package.

Resolves #188

### DIFF
--- a/Package/ExcelDna.AddIn/build/ExcelDna.AddIn.targets
+++ b/Package/ExcelDna.AddIn/build/ExcelDna.AddIn.targets
@@ -69,6 +69,17 @@
   </Target>
 
   <!--
+    Target that ensures ExcelDna.Integration.dll is not copied to the output folder
+    https://github.com/Excel-DNA/ExcelDna/issues/188
+  -->
+  <Target Name="ExcelDnaPreventIntegrationCopyLocal" AfterTargets="ResolveAssemblyReferences">
+    <ItemGroup>
+      <ReferenceCopyLocalPaths Remove="@(ReferenceCopyLocalPaths)"
+                               Condition="'%(ReferenceCopyLocalPaths.Filename)%(ReferenceCopyLocalPaths.Extension)' == 'ExcelDna.Integration.dll'" />
+    </ItemGroup>
+  </Target>
+
+  <!--
     Configure debugger options (Path for EXCEL.EXE, add-in to open, etc).
   -->
   <Target Name="ExcelDnaSetDebuggerOptions" BeforeTargets="PreBuildEvent" Condition="$(RunExcelDnaSetDebuggerOptions) AND '$(BuildingInsideVisualStudio)' == 'true'">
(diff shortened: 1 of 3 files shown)


govert:
Oh cool = thank you @augustoproiete 

Where does this leave us for new-style projects that target .NET Framework?

augustoproiete:
@govert #193 is the last piece of the puzzle now

govert:
@augustoproiete OK great. I want to make a release of v 1.1 soonish - in the next week or so. Is there anything else you want to do before that?

augustoproiete:
@govert Awesome. I might be able to sneak in one more minor improvements to the build in the next couple of days

PR #327, opened 2020-06-28
Allow devs to use PackageReference if they wish (despite current limitations)

augustoproiete:
- Add new build parameter `ExcelDnaAllowPackageReferenceProjectStyle` to the `.props` file (defaults to `false`) to allow for displaying error when users install `ExcelDna.AddIn` on projects that use the `PackageReference` style
- Changing the parameter to `true` allows the developer to bypass the error, if the wish to use `PackageReference` with a workaround for copying the files, etc.

### DIFF
--- a/Package/ExcelDna.AddIn/build/ExcelDna.AddIn.targets
+++ b/Package/ExcelDna.AddIn/build/ExcelDna.AddIn.targets
@@ -54,6 +54,8 @@
     <ExcelDna32BitAddInSuffix Condition="'$(ExcelDna32BitAddInSuffix)' == ''"></ExcelDna32BitAddInSuffix>
     <ExcelDna64BitAddInSuffix Condition="'$(ExcelDna64BitAddInSuffix)' == ''">64</ExcelDna64BitAddInSuffix>
 
+    <ExcelDnaAllowPackageReferenceProjectStyle Condition="'$(ExcelDnaAllowPackageReferenceProjectStyle)' == ''">false</ExcelDnaAllowPackageReferenceProjectStyle>
+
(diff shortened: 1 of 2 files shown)


govert:
Which of these do you prefer for the pre-release NuGet package with this in?
1.1.0-SdkStyleProjects
1.1.0-PackageReference

augustoproiete:
I think `1.1.0-SdkStyleProjects` might be more explicit.

However, I think a pre-release is not very useful as it'll block devs from shipping RTM versions of their own libraries. I've replied to your e-mail just now.

govert:
I'm confused about whether it's the Sdk-style project file, or the NuGet PackageReference that's actuall the problem.
(Or is it the NetStandard target?)

Maybe the best name is 
1.1.0-NetStandard

augustoproiete:
@govert It's both... Changing the old csproj to use PackageReference breaks the current build. The new SDK-style project also breaks the current build.

I think the `NetStandard` is confusing because it implies support for .NET Standard - which we don't support.

govert:
I guess it's the PackageReference reference that doesn't run the Install.ps1 which is our problem. The SdkStyle file just forces that.

So I'm tempted to go with 
1.1.0-PackageReference
maybe?

But I'll sleep on it and push the new package in the morning.

augustoproiete:
`install.ps1` is one of the problems... The `SetDebuggerOptions` also doesn't work in SDK-project style, which means no auto F5 debugging experience anymore (find the right Excel version, etc.)

Now that this one is merged, I suggest releasing `v1.1.1` or `v1.2.0` (RTM) and de-listing `v1.1.0` from NuGet. That avoids any confusion, and gives an RTM version that allows using ExcelDna with SDK-style projects (as long as the parameter is set in the `.props` file)

govert:
Great - yes. I'll test it a bit then make a release as package v 1.1.1.

augustoproiete:
Awesome. Thanks!

PR #334, opened 2020-07-05
Update all projects to target .NET Framework 4.5

augustoproiete:
- C# projects updated to target .NET Framework 4.5
- C++ project updated to use VS2019 build tools (v142)

### DIFF
--- a/Source/ExcelDnaPack/ExcelDnaPack.csproj
+++ b/Source/ExcelDnaPack/ExcelDnaPack.csproj
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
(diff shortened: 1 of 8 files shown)


Source/ExcelDnaPack/ExcelDnaPack.csproj
```diff
@@ -32,6 +32,7 @@
     <LangVersion>3</LangVersion>
     <PlatformTarget>x86</PlatformTarget>
     <UseVSHostingProcess>true</UseVSHostingProcess>
+    <Prefer32Bit>false</Prefer32Bit>
```
govert, line 35: This will change ExcelDnaPack from running as 32-bit to running as 64-bit in most cases. I'm not sure we want to do that, even though it should be fine. Eventually we want it to run as an MSBuild task, which I think normally runs as a 32-bit process, though I'm not sure. At least it needs some thought.

govert:
If we really target 4.5 with the ExcelDna core libraries (which I'm not not sure is needed or useful now, but I'm OK with over the longer term) then we should at least get something like this in place before a release: https://github.com/Excel-DNA/ExcelDna/issues/335

Otherwise there is a danger of using some 4.5+ feature in the Excel-DNA core libraries, and finding a surprising crash at runtime if someone loads under 4.0. I am not really in a position to test stuff under older runtimes, so I rely on the compiler etc. to help me.

Source/ExcelDnaPack/ExcelDnaPack.csproj
govert: In particular, since it calls the Win32 API a bit, I have not checked whether there are 32-bit / 64-bit surprises there.

augustoproiete: Makes sense to test this out then.

govert: Sorry - I only see the `<PlatformTarget>x86</PlatformTarget>` now - I think that means `<Prefer32Bit>false</Prefer32Bit>` is irrelevant.

PR #336, opened 2020-07-05
Set PlatformTarget to x86 for Release builds & also Prefer32Bit for both Debug and Release

augustoproiete:
I noticed only `Debug` builds had the `PlatformTarget` set to `x86`.

Also setting `Prefer32Bit` to `true` for consistency.

Once we add ExcelDnaPack as a build task, we can add test coverage for x64 and make sure it works as expected.


govert:
Great - thanks. I'm not sure this even needs to run as 64-bit.

PR #361, opened 2021-01-04
Remove developmentDependency from ExcelDna.AddIn nuget spec

augustoproiete:
`PackageReference` excludes assembly references by default unfortunately, making `developmentDependency` useless for us given that we need the `ExcelDna.Integration` assembly referenced

https://github.com/NuGet/Home/wiki/DevelopmentDependency-support-for-PackageReference


govert:
Thank you @augustoproiete !

PR #369, opened 2021-02-19
Update to Sdk-style project files

govert:
This moves to Sdk-style project files for the main Excel-DNA C# projects, targeting .NET Framework 4.5.
The test projects have not been tested yet.
Otherwise it should work, but needs some more testing.


Source/ExcelDna.AddIn.Tasks/ExcelDna.AddIn.Tasks.csproj
```diff
@@ -1,72 +1,39 @@
...
-    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <ProjectGuid>{B5CC04D3-0507-4431-9F45-7C60BAC3047A}</ProjectGuid>
+    
     <OutputType>Library</OutputType>
-    <AppDesignerFolder>Properties</AppDesignerFolder>
-    <RootNamespace>ExcelDna.AddIn.Tasks</RootNamespace>
-    <AssemblyName>ExcelDna.AddIn.Tasks</AssemblyName>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
-    <FileAlignment>512</FileAlignment>
+    <TargetFramework>net452</TargetFramework>
```
augustoproiete, line 6:
```suggestion
    <TargetFramework>net472</TargetFramework>
```

I suggest we go `net472` here given this is library used during build time only. Both Visual Studio 2017 and 2019 require .NET Framework 4.7.2 in order to **run**, so we can expect all Excel-DNA developers to have at least `net472` on their development machines.

This should also resolve the warning on `EnvDTE` being restored as `net461`.

Source/ExcelDna.AddIn.Tasks/ExcelDna.AddIn.Tasks.csproj
```diff
@@ -1,72 +1,39 @@
...
+
   </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\ExcelDna.Integration\ExcelDna.Integration.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="EnvDTE" Version="8.0.2" />
+    <PackageReference Include="NUnit" Version="*" />
```
augustoproiete, line 27:
```suggestion
```

I believe this `NUnit` package reference should be removed.

Source/ExcelDna.AddIn.Tasks/ExcelDna.AddIn.Tasks.csproj
```diff
@@ -77,13 +44,7 @@
       <SubType>Designer</SubType>
     </None>
   </ItemGroup>
-  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+  <!--<Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />-->
```
augustoproiete, line 47: This can be safely deleted.

Source/ExcelDna.AddIn.Tasks/ExcelDna.AddIn.Tasks.csproj
```diff
@@ -1,72 +1,39 @@
...
-    <Compile Include="Properties\AssemblyInfo.cs" />
-    <Compile Include="Utils\Bitness.cs" />
-    <Compile Include="Utils\DevToolsEnvironment.cs" />
-    <Compile Include="Utils\ExcelDetector.cs" />
-    <Compile Include="Utils\ExcelDnaPhysicalFileSystem.cs" />
-    <Compile Include="Utils\IExcelDetector.cs" />
-    <Compile Include="Utils\IExcelDnaFileSystem.cs" />
-  </ItemGroup>
+
+  <!-- TODO: Not sure what to do here -->
```
augustoproiete, line 36: Is this comment about the links to the `.props` and `.targets` files? Is yes, then they look good to me. I created these links just to have a shortcut to edit these files directly from Visual Studio.

If the comment is about all the `<Compile ...>` that were removed, they also look good.

(_for later_) Now that this project is converted to SDK-style, we could prob. use MSBuild to pack and get rid of the `.nuspec` (perhaps even `dotnet pack` would work) so something we can consider for a future PR - in which cases the `.props` and `.targets` would need a few more attributes to be included in the package.

Source/ExcelDna.Integration/ExcelDna.Integration.csproj
```diff
@@ -1,120 +1,21 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+﻿<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
```
augustoproiete, line 1:
```suggestion
<Project Sdk="Microsoft.NET.Sdk">
```

We should be able to use the regular .NET SDK here (you're already referencing WinForms) down below.

Source/Tests/ExcelDna.AddIn.Tasks.IntegrationTests/ExcelDna.AddIn.Tasks.IntegrationTests.csproj
```diff
@@ -1,84 +1,16 @@
...
-    <AppDesignerFolder>Properties</AppDesignerFolder>
-    <RootNamespace>ExcelDna.AddIn.Tasks.IntegrationTests</RootNamespace>
-    <AssemblyName>ExcelDna.AddIn.Tasks.IntegrationTests</AssemblyName>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
-    <FileAlignment>512</FileAlignment>
-    <RestorePackages>true</RestorePackages>
-    <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\..\</SolutionDir>
-    <TargetFrameworkProfile />
+    
+    <TargetFramework>net452</TargetFramework>
```
augustoproiete, line 5:
```suggestion
    <TargetFramework>net472</TargetFramework>
```

Similar to above, we can go `net472` here, or multi-target

Source/Tests/ExcelDna.AddIn.Tasks.IntegrationTests/ExcelDna.AddIn.Tasks.IntegrationTests.csproj
```diff
@@ -1,84 +1,16 @@
...
-    <Compile Include="SingleDnaFileAppConfigLinkedIntegrationTests.cs" />
-  </ItemGroup>
-  <ItemGroup>
-    <None Include="packages.config" />
-  </ItemGroup>
+
   <ItemGroup>
-    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
+    <PackageReference Include="ExcelDna.Interop" Version="14.0.1" />
+    <PackageReference Include="NUnit" Version="*" />
```
augustoproiete, line 14: Similar to the above, I believe this `NUnit` reference should be deleted.

Source/Tests/ExcelDna.IntegrationTests/ExcelDna.IntegrationTests.csproj
```diff
@@ -1,58 +1,36 @@
...
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
-    <DebugType>pdbonly</DebugType>
-    <Optimize>true</Optimize>
-    <OutputPath>bin\Release\</OutputPath>
-    <DefineConstants>TRACE</DefineConstants>
-    <ErrorReport>prompt</ErrorReport>
-    <WarningLevel>4</WarningLevel>
+    
+    <TargetFramework>net452</TargetFramework>
```
augustoproiete, line 5:
```suggestion
    <TargetFramework>net472</TargetFramework>
```

Similar to above, we can go `net472` here, or multi-target

Source/Tests/ExcelDna.IntegrationTests/ExcelDna.IntegrationTests.csproj
```diff
@@ -1,58 +1,36 @@
...
+      <Isolated>false</Isolated>
+      <EmbedInteropTypes>true</EmbedInteropTypes>
+    </COMReference>
   </ItemGroup>
+
   <ItemGroup>
-    <Compile Include="Tests.cs" />
-    <Compile Include="MarshalingTests.cs" />
-    <Compile Include="Properties\AssemblyInfo.cs" />
+    <PackageReference Include="NUnit" Version="*" />
```
augustoproiete, line 27: Similar to the above -> :x:

Source/Tests/ExcelDna.IntegrationTests/ExcelDna.IntegrationTests.csproj
```diff
@@ -1,58 +1,36 @@
...
-    </Reference>
-    <Reference Include="System" />
-    <Reference Include="System.Core" />
-    <Reference Include="System.Xml.Linq" />
-    <Reference Include="System.Data.DataSetExtensions" />
-    <Reference Include="Microsoft.CSharp" />
-    <Reference Include="System.Data" />
-    <Reference Include="System.Net.Http" />
-    <Reference Include="System.Xml" />
+    <COMReference Include="Microsoft.Office.Excel.dll">
```
augustoproiete, line 15: Is this required? Could we replace this `COMReference` with `<PackageReference Include="ExcelDna.Interop" Version="14.0.1" />` ?

Source/Tests/ExcelDna.IntegrationTests/ExcelDna.IntegrationTests.csproj
```diff
@@ -63,30 +41,17 @@
...
-  <PropertyGroup>
-    <PostBuildEvent>
-    </PostBuildEvent>
-  </PropertyGroup>
-  <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
-       Other similar extension points exist, see Microsoft.Common.targets.
-  <Target Name="BeforeBuild">
-  </Target>
-  -->
+  </ItemGroup>-->
```
augustoproiete, line 51: This can be safely deleted.

augustoproiete (approved):
LGTM! Some minor comments throughout.

Big step forward and opens a lot of doors! :rocket:

Once this is merged, I can follow-up with PRs to add deterministic builds, source link, and some general build versioning / packaging attributes

Source/ExcelDna.Integration/ExcelDna.Integration.csproj
govert: I think this is still needed when targeting the framework. It's kind of a bug, and they're fixing for some point version of .NET 5 SDK.

govert:
@augustoproiete Thank you very much for looking at this!
Yes - I think the next big step would be to try to build the NuGet packages from the project files.

Source/ExcelDna.AddIn.Tasks/ExcelDna.AddIn.Tasks.csproj
govert: It looks like this is used in the attributes on the test methods. So keeping for now.

Source/Tests/ExcelDna.AddIn.Tasks.IntegrationTests/ExcelDna.AddIn.Tasks.IntegrationTests.csproj
govert: Also used - keeping for now.

Source/Tests/ExcelDna.IntegrationTests/ExcelDna.IntegrationTests.csproj
govert: This breaks when changing to a PackageReference to `ExcelDna.Interop`. I presume `ExcelDna.Interop` needs some work to allow embedding with the PackageReference mechanism.

Issue #376, opened 2021-03-24
IsInFunctionWizard worry


ittegrat:
**Problem**
The '**Edit Links**' dialog is mistakenly detected as the Function Wizard dialog. The manual update of external links that trigger the calculation of functions that use **ExcelDnaUtil.IsInFunctionWizard** could lead to wrong results.

**Details**
The class name of the 'Edit Links' dialog start with 'bosa_sdm_XL'. The obvious patch would be to add a check `title.Contains("Links")` [here](https://github.com/Excel-DNA/ExcelDna/blob/082d3d03a7068389d7016dd31e1b1e7a7f988e62/Source/ExcelDna.Integration/Excel.cs#L505) but I was wondering if it wouldn't be better to have a positive logic like:
```
if (title.StartsWith("Function Arguments"))
  return true;
else
  return false;
```
inside the **IsFunctionWizardWindow** delegate.

Unfortunately I only have access to a recent version of Excel 365 (v2002) and I can't verify the title of the dialog for older versions of Excel.

In either case, I could provide a PR.

Regards

govert:
Checking the title text would be problematic for localised versions.
I think one would need some check in the structure of the dialog - not sure.

What is the problem when a function does not calculate while Edit Links is shown?

govert:
I see the code already checks for title names, so under localised versions would go wrong, but I guess in the 'safe' direction.

ittegrat:
As an example, consider a function that use the check to skip a lengthy calculation. You open the dialog, update the inputs, the function skip the recalculation and you are forced to trigger a full recalculation after closing the dialog. I assume that the check for 'Replace' contained in the dialog title has been introduced to resolve the same issue with the 'Find and Replace' dialog.

Pass the title to be checked (or an array of tiltles) as input to the IsInFunctionWizard function, so it is developer responsibility to provide the right string, could be an option ?

govert:
To do it better one might have to look at the window classes of child windows of the particular  'bosa_sdm_XL' and try to detect it that way. It's a bit tedious to dig into, but maybe you're interested?

Otherwise, I think my preference for now would be to just add the `title.Contains("Links")` check, which works in English at least.
All the code needed to implement this independently is about 20 lines anyway, so one could make a separate sample if others need to customize with localized titles etc.

govert:
@ittegrat Thank you for the contribution. Your update is in version 1.2.5-preview which I've pushed here and on GitHub.

PR #377, opened 2021-03-26
Check child window classes to detect the Function Wizard dialog

ittegrat:
As suggested in #376, here a children check to detect the right dialog.
I've managed to test a C++ equivalent method on an old Excel 2003 / Win7 installation, so I presume this solution should be backward compatible.


govert:
OK great - I'll give it a try.
It would be funny if they now change the function dialog after all these decades :-)

PR #381, opened 2021-04-24
Addin.Task is compatible with net5.0-windows

PierreYvesR:
### Summary
Projects that use ExcelDna.Addin can be built with dotnet core 5.0 on windows directly through ```dotnet build```, so thereby not requiring Visual Studio to be installed.

### Description
Multi-target Addin.Tasks for net472 and net5.0-windows.
Adds a new variable ExcelDnaTasksPath to the Excel.Addin.Targets to redirect to the correct Excel.Addin.Tasks.dll during build.

It's my first PR on ExcelDna - so consider this a draft PR...


govert:
Thank you Pierre-Yves - that sounds very promising. I'll be able to have a
look next week.
-Govert

PR #387, opened 2021-07-06
Some simplifications of the C++ project build

davidhunter22:
This PR tried to simplify the build of the C++ project. There are three compoents
1) Remove solution platforms that are not needed - This leaves just x86 and Win32 -  Note C# still builds using a project plarform of AnyCPU as before
2) Simplify the vcxproj file, see check in comment for more details
3) Use the build in assember functionality from MSBuild rather than calling the assembler directly

All these changes should be independent, let me know if you would rather split them up.


govert:
Thank you @davidhunter22 !
Do you know off-hand what goes wrong when the 32-bit platform is called 'x86' instead of 'Win32' ?

davidhunter22:
Last time I tried x86 as a project platform, which was a while back, it didn't work. The props and target files in places like 
C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/MSBuild/Microsoft/VC/v160 have Win32 hard coded and don't seem to also accept x86 as an alternative spelling.

You can change the solution platform names, even to something like "Excel32" and "Excel64". 
You can also change the output directory to be x86. We did that on a few projects as it matches what C# SDK style project s prefers.

Note that if you use project platforms like x86 and x64 on C# projects the only thing it changes in the output assembly is the metadata architecture which you can see in ILSpy. so x86 will cause it to be 32 bit preferred. This can be useful say on a 64 bit machine if you want to force the CLR to run 32 bit.

Issue #396, opened 2021-07-19
Use Dot Net Core SDK direct from the install tree rather than a copy.


davidhunter22:
There seems to be a copy of the 32 bit libnethost.lib and header files inside ExcelDna.Host. An alternative would be to reference them from the SDK install tree. I don't think it's unreasonable to say that installing the SDk is a prerequisite of building ExcelDNA itself. I will create a PR so you can take a look, this will include getting the 64 bit build of ExcelDna.Host to work as well.

You mentioned dealing with the packed ExcelDna.ManagedHost.dll file. When you say "dealing with" do you mean loading into  the CLR in the  ExcelDna.Host code?

I do have otehr build suggestions, the following are roughly in order of which I think might be most impactful.

1) Generate the nuget packages as part of the csproj files rather than using nuspec files and batch commands, this uses <GeneratePackageOnBuild>. This way the packages generation is simply part of the normal build process with no other messing around. This could also include symbols via <IncludeSymbols>. Note this is orthogonal to the issues of users consuming the nuget package via <PackageReference> rather than packages.config. 
2) Factor out more build stuff into Directory.Build.props
3) Put all build output into a common solution wide output directory rather than into each project and then copying stuff around
4) Select a C# compiler version, via <LangVersion>, I would suggest 9.0
5) Switch on nullabel reference types.
6) use /W4 and treat warnings as erros.
7) Use static analysis tools

govert:
I think it's simpler to keep a separate copy of the nethost files. There is a way to reference a C++ package to get them, but that seems more complicated. I'm trying to keep stuff less complicated for now. Referencing from the SDK has all kinds of version issues.

I suggest we not fiddle too much with the build stuff for now, please. I definitely can't handle large-scale changes to the project (at least for now), so I think you'll be wasting your time there - sorry. And I don't want to refactor the managed code right now, unless we need to make functional changes. The managed code mostly needs to run under .NET 4.5.2 too.

There is a good opportunity to help with the .NET 5+ hosting, since that's not baked at all yet. 
I've created a good issue to start with here: https://github.com/Excel-DNA/ExcelDna/issues/398

govert:
Another issue that can do with help without affecting the existing stuff is the whole PackageReference story: https://github.com/Excel-DNA/ExcelDna/issues/363

I don't know exactly how that should work yet, but you might have some suggestions based on my first dump of ideas in that issue.

[...]

PR #397, opened 2021-07-19
Use dot net core SDK from install

davidhunter22:
This is a possible implementation of issue  #396


govert:
I'm not keen on this handling of the nethost because of the version issues.

PR #423, opened 2022-01-24
Added graceful SynchronizationManager shutdown on ExcelComAddIn.OnDisconnection()

Sergey-Vlasov:
To prevent LPenHelper AV on Excel shutdown when ExcelComAddIn is used.

### DIFF
--- a/Source/ExcelDna.Integration/ExcelComAddIn.cs
+++ b/Source/ExcelDna.Integration/ExcelComAddIn.cs
@@ -47,6 +47,7 @@ namespace ExcelDna.Integration
 
         public virtual void OnDisconnection(ext_DisconnectMode RemoveMode, ref Array custom)
         {
+            SynchronizationManager.Uninstall();
             Logger.ComAddIn.Verbose("ExcelComAddIn.OnDisconnection");
         }
 
--- a/Source/ExcelDna.Integration/ExcelSynchronizationContext.cs
+++ b/Source/ExcelDna.Integration/ExcelSynchronizationContext.cs
@@ -88,7 +88,7 @@ namespace ExcelDna.Integration
         {
             get
             {
-                if (_syncWindow != null && _syncWindow.RunMacroSynchronization.IsRegistered) 
+                if (_syncWindow != null && _syncWindow.RunMacroSynchronization.IsRegistered)
                     return _syncWindow.RunMacroSynchronization;
 
                 return null;
@@ -343,7 +343,7 @@ namespace ExcelDna.Integration
 
             object xlCallResult;
             XlCall.TryExcel(XlCall.xlfRegister, out xlCallResult, registerParameters);
-            Logger.Registration.Verbose("Register SyncMacro - XllPath={0}, ProcName={1}, FunctionType={2}, MethodName={3} - Result={4}", 
+            Logger.Registration.Verbose("Register SyncMacro - XllPath={0}, ProcName={1}, FunctionType={2}, MethodName={3} - Result={4}",
                 registerParameters[0], registerParameters[1], registerParameters[2], registerParameters[3], xlCallResult);
             if (xlCallResult is double)
             {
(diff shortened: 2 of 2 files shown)


Source/ExcelDna.Integration/ExcelSynchronizationContext.cs
```diff
@@ -363,8 +363,20 @@ public bool IsRegistered
         void Unregister()
         {
             // Clear the name and unregister
-            XlCall.Excel(XlCall.xlfSetName, _syncMacroName);
-            XlCall.Excel(XlCall.xlfUnregister, _syncMacroRegistrationId);
+            try
```
govert, line 366: These calls should rather be XlCall>TryExcel.

Source/ExcelDna.Integration/ExcelComAddIn.cs
```diff
@@ -47,6 +47,7 @@ public virtual void OnConnection(object Application, ext_ConnectMode ConnectMode
 
         public virtual void OnDisconnection(ext_DisconnectMode RemoveMode, ref Array custom)
         {
+            SynchronizationManager.Uninstall();
```
govert, line 50: Should this rather be called in OnBeginShutdown? We might have different ComAddIn objects loaded, and they might be unloaded without Excel closing while we want the rest of the add-in to keep working. 
We're basically trying to use whatever ExcelComAddIn object is created (for a ribbon, as a CTP helper or just by the user) and use that to get an Excel Shutdown signal. The user might override the event ExcelComAddIn handler in their code and not call the base method, or there might be multiple such objects running in a session.

govert: Maybe we can also add a comment explaining this in the code.

govert (changes requested):

Source/ExcelDna.Integration/ExcelComAddIn.cs
Sergey-Vlasov: Can you give an example of an ComAddIn object unloaded without Excel closing, so I could experiment?

Source/ExcelDna.Integration/ExcelSynchronizationContext.cs
Sergey-Vlasov: I've now found that if I open Excel without opening a workbook, with ExcelDna.Test add-in, it fails on closing Excel with AccessViolationException on the Excel12v call line at 

> 	ExcelDna.Loader!ExcelDna.Loader.XlCallImpl.TryExcelImpl12(int xlFunction, out object result, object[] parameters) Line 123	C#
>  	ExcelDna.Loader!ExcelDna.Loader.XlCallImpl.TryExcelImpl(int xlFunction, out object result, object[] parameters) Line 75	C#
>  	ExcelDna.Loader!ExcelDna.Loader.IntegrationHost.TryExcelImpl(int xlFunction, out object result, object[] parameters) Line 31	C#
>  	ExcelDna.Integration.dll!ExcelDna.Integration.ExcelIntegration.TryExcelImpl(int xlFunction, out object result, object[] parameters) Line 32	C#
>  	ExcelDna.Integration.dll!ExcelDna.Integration.XlCall.TryExcel(int xlFunction, out object result, object[] parameters) Line 1111	C#
>  	ExcelDna.Integration.dll!ExcelDna.Integration.XlCall.Excel(int xlFunction, object[] parameters) Line 1094	C#
>  	ExcelDna.Integration.dll!ExcelDna.Integration.RunMacroSynchronization.Unregister() Line 368	C#
>  	ExcelDna.Integration.dll!ExcelDna.Integration.RunMacroSynchronization.Dispose() Line 385	C#
>  	ExcelDna.Integration.dll!ExcelDna.Integration.SynchronizationWindow.Dispose() Line 729	C#
>  	ExcelDna.Integration.dll!ExcelDna.Integration.SynchronizationManager.Uninstall() Line 66	C#
>    ExcelDna.Integration.dll!ExcelDna.Integration.ExcelComAddIn.OnDisconnection(ExcelDna.Integration.Extensibility.ext_DisconnectMode RemoveMode, ref System.Array custom) Line 50	C#

So, on shutdown we probably shouldn't call Excel at all.

govert:
Yes - I think we should not make these changes. There is another approach to prevent the LPenHelper calls, which I take in the IntelliSense, to just add a flag when shutdown is detected, and then skip the LPenHelper call after the flag is set. But I thin kwe need the AccessViolation check from the other Pull Request anyway, and then we are OK. We can revisit if there are more problems we run into.

Sergey-Vlasov:
> I think we should not make these changes

OK

PR #424, opened 2022-01-24
Implemented LPenHelper invocation from native code

Sergey-Vlasov:
With the handler for all exceptions.


keithalewis:
This is not used by Excel. It is an artifact from the days of Windows for Pen Computing running on Windows 3.1.
https://winworldpc.com/product/windows-for-pen-comp/1x

govert:
@keithalewis It's a critical part of the magic I use for the IntelliSense, and also a fast and (mostly) reliable way of knowing whether the user is editing a formula. For .NET 6 we sometimes have to jump through an extra hoop though.

keithalewis:
@govert I'm not even going to ask how you figured that out. You are the man!

PR #431, opened 2022-02-15
Optionally disable ExcelDnaPack compression

Phundamentals:
My first pull request ever. Hope this goes well :-)

I've had some trouble with virus scanners since around october 2021. Read some stories about an increase of xlls spreading virusses as well, so they might have had to become more stringent.

I've had success disabling the compression feature of ExcelDnaPack. The scanner, in this case Thor Agent from Heimdal security, does not block my xll anymore.

This change makes the possibility of disabling compression more visible. Hope it helps somebody.

### DIFF
--- a/Package/ExcelDna.AddIn/content/ExcelDna.Build.props
+++ b/Package/ExcelDna.AddIn/content/ExcelDna.Build.props
@@ -72,6 +72,15 @@
       Suffix used for packed .xll files e.g. MyAddIn-packed.xll
     -->
     <ExcelDnaPackXllSuffix Condition="'$(ExcelDnaPackXllSuffix)' == ''">-packed</ExcelDnaPackXllSuffix>
+
+    <!--
+      Options used when packing .xll files.
+      When in trouble with your virus scanner, try uncommenting at least the resource compression properties.
+      The default is true for both compression and multithreaded runs when not specified.
+    -->
+    <!--<ExcelDnaPackCompressResources Condition="'$(ConfigurationName)' == 'Debug'">false</ExcelDnaPackCompressResources>-->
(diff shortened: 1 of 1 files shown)


govert:
Thank you @Phundamentals  - I'm a bit surprised that you're having success against the anti-virus false positives when you say that not compressing the packed assembly works better. For 
v1.6.0 I've added a simple encoding step that seemed to help me. At the moment there is no flag to switch the encoding on or off when packing (it's baked into the native code as well, but one could make a plan). Any feedback on your experience with that would also be very welcome.

Phundamentals:
After some research I discovered that the packed xll mainly triggered the heuristics / behavioral analysis engine of the scanner. The 'dectection' was classified as HEUR/XXX generic virus. This made me think the xll was not actually recognized as a virus, it just ticked too many boxes: we have an executable, which contains more executables, and it uncompresses them. This must be a virus, let's block it to be sure. In my case disabling the compression helped, but I can't be sure if it will last.
I'll have to look up what you mean by adding an encoding step, not sure if I'll understand. And you're welcome, I very much enjoy using ExcelDna and all the effort you guys have put into it!

garethhayter:
I found a similar issue. I was compressing files in my installer with a slightly customised version of LZMA (7zip) and at some point some anti-virus products started flagging the installer. So I changed to normal zip format and then they could obviously unzip and inspect the contents, and they were happy with that. So, it seems that as long as the anti-virus tools are able to uncompress and inspect, then they are happy.

Phundamentals:
Thank you @garethhayter, that is valuable info.

garethhayter:
@Phundamentals Can this setting be enabled by specifying a setting in the .dna file? How?

Phundamentals:
@garethhayter You can just uncomment lines in Properties\ExcelDna.Build.Props in your project:

`
    
    
    `

E.g. uncommenting the first line will disable resource compression for the DEBUG build/profile in Visual Studio.

PR #472, opened 2022-06-24
Call TlbExp to make a type library that is embedded in the .xll

Sergey-Vlasov:
For .NET 6, dscom need to be installed on the development machine:
https://github.com/dspace-group/dscom

And for .NET 6, a COM interface should be explicitly declared, not using ClassInterfaceType.AutoDual.


govert:
> And for .NET 6, a COM interface should be explicitly declared, not using ClassInterfaceType.AutoDual.
Do you think AutoDispatch still works?

govert:
Yes I think that code can only work if the `IDispatch` interface is implemented. 
I think it will fail if you put `ClassItnerfaceType.None`.

I also think that `AutoDual` still works the same way under .NET 6 as under Framework - it's just that the dscom tool chose not to support it (which I think is fine, for the problematic reasons they point to.)

Sergey-Vlasov:
> I also think that `AutoDual` still works the same way under .NET 6 as under Framework - it's just that the dscom tool chose not to support it (which I think is fine, for the problematic reasons they point to.)

It is not only dscom. In .NET 6 build, I got 
warning CS0618: 'ClassInterfaceType.AutoDual' is obsolete: 'Support for IDispatch may be unavailable in future releases.'

And there is CA1408: Do not use AutoDual ClassInterfaceType
https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1408?view=vs-2022&tabs=csharp

Sergey-Vlasov:
> Yes I think that code can only work if the `IDispatch` interface is implemented. I think it will fail if you put `ClassItnerfaceType.None`.

In NET6DnaComServer/AddIn.cs:

```
    [ComVisible(true)]
    [ClassInterface(ClassInterfaceType.None)]
    public class ComLibrary : IComLibraryInterface
```

govert:
I'm actually super confused. See my answer here: https://groups.google.com/g/exceldna/c/onRDZLJwoBY
I had discovered the issue when making the CTP work under .NET 6. This was the crucial information: https://github.com/dotnet/winforms/issues/4370#issuecomment-798828348
For the UserControl (the context in both the CTP and the WinForms issue above) it does not work to mark the derived class as `AutoDispatch`. You have to implement an interface and mark it as the `ComDefaultInterface`. Apparently this is needed for the `IDispatch` to be created. The way I'm reading it, the `IDispatch` is no longer made using reflection (so is made at compile time in the interface case, so still works?) But then what about your example above? Perhaps something to do with the base class. The `UserControl` class is no longer marked as `ComVisible` in .NET 5/6. So having a base class that is not ComVisible (so was not compiled knowing that it would need to support `IDispatch` might break the `AutoDispatch` by requiring the reflection-based implementation which is no longer supported.

Anyway, it's worth taking note of these quirks, especially about the CTP controls. I forgot about the problem and fix I found in February, just that I had tried it under .NET 6 and did get it to work.

govert:
I think your example:    
```
    [ComVisible(true)]
    [ClassInterface(ClassInterfaceType.None)]
    public class ComLibrary : IComLibraryInterface
```
work because of the interface, where the default is then to generate an IDispatch.

PR #493, opened 2022-08-13
.NET 6 add-in reloading

Sergey-Vlasov:


govert:
Hi @Sergey-Vlasov - can you split the first commit (fixing the reloading error) from the further reloading support into separate pull requests? I think I want to keep changes to the ExcelDna.Loader assembly for the next version.

PR #512, opened 2022-10-06
Added sorting for directory entries in ResourceHelperManaged

Sergey-Vlasov:
Fixed Linux packed file does not work for project names starting with a letter before E.


govert:
Thank you @Sergey-Vlasov :-)

PR #540, opened 2022-12-18
Added packing native libraries from .deps.json and resolving them at runtime

Sergey-Vlasov:
1. Increased ExcelDna.AddIn.Tasks project's target framework from net452 to net462.

2. I couldn't find a way to clean the temp directory as there are no unloading events available in ExcelDna.ManagedHost.

### DIFF
--- a/Build/build.bat
+++ b/Build/build.bat
@@ -21,6 +21,7 @@ if not exist "..\Package\ExcelDna.AddIn\tools\net6.0-windows\" mkdir "..\Package
 copy /Y ..\Source\ExcelDna.AddIn.Tasks\bin\Release\net462\ExcelDna.AddIn.Tasks.dll ..\Package\ExcelDna.AddIn\tools\net452\
 copy /Y ..\Source\ExcelDna.AddIn.Tasks\bin\Release\net462\ExcelDna.AddIn.Tasks.pdb ..\Package\ExcelDna.AddIn\tools\net452\
 copy /Y ..\Source\ExcelDna.AddIn.Tasks\bin\Release\net462\Newtonsoft.Json.dll ..\Package\ExcelDna.AddIn\tools\net452\
+copy /Y ..\Source\ExcelDna.AddIn.Tasks\bin\Release\net462\Microsoft.Extensions.DependencyModel.dll ..\Package\ExcelDna.AddIn\tools\net452\
 copy /Y ..\Source\ExcelDna.AddIn.Tasks\bin\Release\net6.0-windows\ExcelDna.AddIn.Tasks.dll ..\Package\ExcelDna.AddIn\tools\net6.0-windows\
 copy /Y ..\Source\ExcelDna.AddIn.Tasks\bin\Release\net6.0-windows\ExcelDna.AddIn.Tasks.pdb ..\Package\ExcelDna.AddIn\tools\net6.0-windows\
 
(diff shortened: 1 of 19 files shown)


govert:
1. Would it make sense to use a stable guid for the temp directory where we extract ExcelDna.ManagedHost.dll and the native libraries? We have a stable guid per add-in, derived from the add-in path, which is used in the IntelliSense. See https://github.com/Excel-DNA/RegistrationInfoGuid and https://github.com/Excel-DNA/ExcelDna/blob/master/Source/ExcelDna.Integration/GuidUtility.cs

2. In addition to the native libraries, I'm hoping we can add the managed platform-specific libraries from the .deps.json file into the ExcelDnaPack list as well. Thus that the example add-in would work without an `<ExcelAddInInclude>` property in the project file. It would select the platform specific version of System.Management for packing in a similar way.  I think the Microsoft.Data.SqlClient package, or an EFCore package, has both native and platform specific managed dependencies. Do you want to add this as part of this PR, or a separate PR and have me pull this first? I'm happy either way.

3. Do you think we need a property or similar to control this behaviour of that packing task and opt out if needed?

govert:
4. Do you think we need a property to exclude specific file(s) from the deps.json-based packing?

Sergey-Vlasov:
1. With a stable guid, it will not be a temp dir, but an add-in storage dir. With a different design and two implementations for C++ and C# code.

Sergey-Vlasov:
2. A separate PR. I expect automatically packing managed libraries will be a major (breaking?) change in user experience for a packed add-in.

Sergey-Vlasov:
3,4. I've added ExcelDnaPackNativeLibraryDependencies and ExcelDnaPackExcludeDependencies project properties.

govert:
I suggest we call it ExcelDnaPackDependencies (with default true) instead of ExcelDnaPackNativeLibraryDependencies, and use that for both the managed and native dependencies.

Sergey-Vlasov:
I’m still uncertain about packing managed dependencies. First, .deps.json is imprecise in specifying managed dependencies. And then it has to be integrated with existing managed packing options. 

I thought about a second ExcelDnaPackManagedDependencies option.

govert:
OK - we can keep two properties then, and see how the managed dependencies work as a separate PR.

The way I think about it is that we if PackxxxDependencies is enabled, then we pack the managed and native dependencies from .deps.json, plus any files in ExcelAddInInclude minus any files in ExcelDnaPackExcludeDependencies.
So that by default an add-in author can add NuGet packages, and build, and the packed file will have everything - managed and native.

Sergey-Vlasov:
Should I set default true for ExcelDnaPackNativeLibraryDependencies?

Sergey-Vlasov:
I've set default true for ExcelDnaPackNativeLibraryDependencies.

govert:
> I couldn't find a way to clean the temp directory as there are no unloading events available in ExcelDna.ManagedHost.

Do I understand right that this means every run of the add-in will create a new directory with the files in?
That's way I asked about the stable directory.

> With a stable guid, it will not be a temp dir, but an add-in storage dir. With a different design and two implementations for C++ and C# code.

I want to see how this works first, so can you change the default back to false for ExcelDnaPackNativeLibraryDependencies?

govert:
> > With a stable guid, it will not be a temp dir, but an add-in storage dir. With a different design and two implementations for C++ and C# code.

> I want to see how this works first, so can you change the default back to false for ExcelDnaPackNativeLibraryDependencies?

I mean I just want to try the current implementation first, not that we should do the storage dir yet.

Sergey-Vlasov:
> Do I understand right that this means every run of the add-in will create a new directory with the files in?

Yes.

Sergey-Vlasov:
> I want to see how this works first, so can you change the default back to false for ExcelDnaPackNativeLibraryDependencies?

Done.

PR #575, opened 2023-02-25
Fix broken documentation link and add bullets

adrianocorrea:
I fixed the broken link to the documentation. 
I am also suggesting adding bullets to improve readability.


govert:
Thanks @adrianocorrea - as you see we're in the process of breaking all the documentation stuff, hoefully towards a better future.

PR #581, opened 2023-03-23
Copy .chm help file from ExcelDnaDoc to publish directory

Sergey-Vlasov:


Source/ExcelDna.AddIn.Tasks/CreateExcelAddIn.cs
```diff
@@ -148,6 +150,21 @@ private void TryCreateTlb()
             }
         }
 
+        private void TryPublishDoc()
+        {
+            string docFile = AddInFileName ?? ProjectName + "-AddIn" + ".chm";
```
govert, line 155: I'm guessing we need some brackets here.

Sergey-Vlasov: Thank you for noticing. I've corrected it.

PR #738, opened 2024-11-16
Fixed the MasterBuild ExcelDnaPack win-x86 target error

Sergey-Vlasov:


govert:
@Sergey-Vlasov This change breaks the command-line build on my machine (without .NET 9 installed) though it works under VS.

Can you check whether this works under .NET 9 - it might give us compatibility for a version of two?

```
    <RuntimeIdentifiers>win7-x86;win7-x64</RuntimeIdentifiers>
    <UseRidGraph>true</UseRidGraph>
```

I don't really understand why .NET 8 and .NET 9 behave differently. It seems to be related to the ExcelDnaPack being built as a dependency of ExcelDna.vcxproj, when it is passed the x64 platform. But the error I see is confusing.
Anyway - I would just like to find a settings that is compatible with both .NET 8 and .NET 9 builds - then we can sort out why we need the `RuntimeIdentifiers` it at all much later. It seems a quirk of the msbuild restore target.

Sergey-Vlasov:
Yes, with `<UseRidGraph>true</UseRidGraph>` it works on my main machine under .NET 9.

PR #770, opened 2025-05-29
Update ExcelDna.AddIn.targets for .net10

valdarly:
Currently checks for TargetFrameworkVersion.Substring(1,1) >= 6 which fails with net10.0. Not sure if the test is .net core vs framework or actually requires .net6.0 or higher. Have updated to using TargetFramework instead of TargetFrameworkVersion. Have implemented very simple .Substring(0,3) == 'net' or not.


valdarly:
Not sure of etiquette here (first time doing a pull request). I have updated the code with your suggestion.

govert:
Thanks @valdarly and @Sergey-Vlasov

PR #802, opened 2025-10-04
Removed WindowsForms dependency for NativeAOT

Sergey-Vlasov:


Source/ExcelDna.Integration/Excel.cs
```diff
@@ -459,7 +463,6 @@ private static object GetApplicationFromWindow(IntPtr hWndMain, bool allowProtec
 
                 // Marshal to .NET, then call .Application
                 object obj = ComInterop.Util.TypeAdapter.GetObject(pUnk);
-                Marshal.Release(pUnk);
```
govert, line 462: Why can we skip the Release here? It seems we end up calling Marshal.GetObjectForIUnknown which increments the reference count.

PR #821, opened 2025-12-22
Invoke ProcessStartupHooks with the correct number of arguments

kosak:
Behavior before this change: 
Every time you start the add-in this line appears in the Debug log: `Exception thrown: 'System.Reflection.TargetParameterCountException' in System.Private.CoreLib.dll`

After this change: that line no longer appears in the Debug log.

Rationale for the change:

The method we are invoking is declared `private static void StartupHookProvider.ProcessStartupHooks(string diagnosticStartupHooks)`

However we are invoking it incorrectly:
when we say `.Invoke(null, null)` the first null means invoke a static method, and the second null means "there are no arguments"

It is incorrect to assert no arguments, because there is in fact one argument: `string diagnosticStartupHooks`

I believe what is intended here is to invoke the method with one argument whose value is null.

So the correct invocation is
``` ProcessStartupHooksMethod.Invoke(null, new object[] { null });```

I have tested this change in my own sandbox, but I haven't tested it in the whole ExcelDNA product because I don't know what this feature is supposed to accomplish and what a successful test looks like. But I can say it takes its intended action now rather than 100% failing.

### DIFF
--- a/Source/ExcelDna.ManagedHost/AddInInitialize.cs
+++ b/Source/ExcelDna.ManagedHost/AddInInitialize.cs
@@ -107,7 +107,7 @@ namespace ExcelDna.ManagedHost
             {
                 Type StartupHookProviderType = Type.GetType($"System.StartupHookProvider, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e");
                 MethodInfo ProcessStartupHooksMethod = StartupHookProviderType.GetMethod("ProcessStartupHooks", BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.InvokeMethod);
-                ProcessStartupHooksMethod.Invoke(null, null);
+                ProcessStartupHooksMethod.Invoke(null, new object[] { null });
             }
             catch
             {

### CODE REVIEW

**Comment by govert** [comment]
@kosak Thank you for looking at this and your contribution.
It seems like the signature of that method changed some time after .NET 6 (at least for .NET 8).
For now, I'd prefer to continue supporting .NET 6+ versions with the same Excel-DNA build.

@Sergey-Vlasov Do you know why we want to call this? Is it just recommended for a new ALC? I can't see clear documentation.

ChatGPT suggested a defensive implementation, that also does not refer to the assembly version.

```
using System;
using System.Linq;
using System.Reflection;

var corelib = typeof(object).Assembly;
var t = corelib.GetType("System.StartupHookProvider", throwOnError: false);
if (t == null) throw new InvalidOperationException("System.StartupHookProvider not found.");

var methods = t.GetMethods(BindingFlags.NonPublic | BindingFlags.Static)
               .Where(m => m.Name == "ProcessStartupHooks")
               .ToArray();

// Prefer ProcessStartupHooks(string) if it exists, else fallback to ProcessStartupHooks()
var m = methods.FirstOrDefault(x =>
            x.GetParameters().Length == 1 &&
            x.GetParameters()[0].ParameterType == typeof(string))
        ?? methods.FirstOrDefault(x => x.GetParameters().Length == 0);

if (m == null) throw new MissingMethodException(t.FullName, "ProcessStartupHooks");

object?[] args = (m.GetParameters().Length == 1)
    ? new object?[] { null }   // or "" depending on desired behavior
    : Array.Empty<object?>();

m.Invoke(null, args);
```

**Comment by kosak** [comment]
Thanks @govert. Good catch, I didn't think to go back and test it in .NET 6.0.

I think I can improve further on ChatGPT's defensive implementation with the below. This gets rid of the LINQ searches and to my eye exposes the intent more clearly. (I also pass 'true' to GetType to let it throw the error rather than having the code duplicatively do it. On the other hand, I agree with ChatGPT that `BindingFlags.InvokeMethod` isn't needed.)

But the point is (to state the obvious) both ChatGPT's code and this one have a behavioral difference from the original: the original swallows all exceptions, but these replacements expose them.

```
        public static void ProcessStartupHooks() {
            var corelib = typeof(object).Assembly;
            var t = corelib.GetType("System.StartupHookProvider", throwOnError: true);

            // Starting with .NET 8.0, System.StartupHookProvider.ProcessStartupHooks takes a single string argument.
            MethodInfo m = t.GetMethod("ProcessStartupHooks", BindingFlags.NonPublic | BindingFlags.Static,
                null, new[] { typeof(string) }, null);
            if (m != null) {
                m.Invoke(null, new object[] { null });
                return;
            }

            // Prior to NET 8.0, System.StartupHookProvider.ProcessStartupHooks takes no arguments
            m = t.GetMethod("ProcessStartupHooks", BindingFlags.NonPublic | BindingFlags.Static,
                null, Array.Empty<Type>(), null);
            if (m != null) {
                m.Invoke(null, null);
                return;
            }

            throw new MissingMethodException(t.FullName, "ProcessStartupHooks");
        }

```

Sergey-Vlasov:
I've created a test project https://github.com/Excel-DNA/ExcelDna/tree/StartupHooks/Source/Tests/ExcelDna.AddIn.Tasks.IntegrationTests.TestTarget/StartupHooks

It confirms that the existing code works for .NET 6/7, but fails for .NET 8/9/10.

PR #839, opened 2026-03-15
Added stable guids for ExcelCustomTaskPaneAddIn

### LINKED ISSUE #373: ProgId/Guid for Custom Task Pane Helper
**Problem**
There is no easy way to set ProgId/Guid for a Custom Task Pane Helper component (`ExcelCustomTaskPaneAddIn`), which might cause a problem in restricted environments.

**Details**
`ExcelCustomTaskPaneAddIn` is an internal class and the instance is being created in the `GetCTPFactory` method inside `CustomTaskPaneFactory`:
```
// Register and create addin
_addin = new ExcelCustomTaskPaneAddIn { DnaLibrary = DnaLibrary.CurrentLibrary };
ExcelComAddInHelper.LoadComAddIn(_addin);
```
`LoadComAddIn` checks if the class has ProgId/Guid attribute, and if not it generates id based on the xll path. It works well if you e.g.  extend `ExcelRibbon` class, but `ExcelCustomTaskPaneAddIn` is an internal class, so you cannot do much with it, therefore it always uses generated id.

Sergey-Vlasov:


govert:
Can we really hardcode the guid and progid we use for the ctp add-in?

govert:
I mean how does that interact when there are multiple add-ins - are we sure these will only be used in the temporary registration? I think the point is to be able to have per-addin guid and progid that can be pre-registered with the corporate security software, allowing those add-ins to load. That might mean it should be easier to bring the CTP loader add-in into the add-in itself, and not have to use the Excel-DNA provided loader helper only.

Sergey-Vlasov:
The ExcelCustomTaskPaneAddIn is registered until Excel is closed. With the hardcoded guids and multiple add-ins, the second add-in fails to create CTP.

I’ve added support for custom CTP add-in guids instead of the hardcoded ones. With the new CreateCustomTaskPane overloads, it is now possible to have per-addin guid and progid that can be pre-registered with the corporate security software.

PR #840, opened 2026-03-17
add NAOT support for excel functions w 16-29 arguments

mozcanaurelius:
Currently, AOT compilation doesn't support Excel functions with more than 16 arguments, I've added additional delegate types and changed the generation / function registration flows to make it work.

### DIFF
--- a/Source/ExcelDna.Integration/Registration/AotExtendedFunc.cs
+++ b/Source/ExcelDna.Integration/Registration/AotExtendedFunc.cs
@@ -1,0 +1,815 @@
+using System;
+
+namespace ExcelDna.Registration
+{
+    public delegate TResult AotFunc17<
+        T1,
+        T2,
+        T3,
+        T4,
+        T5,
+        T6,
+        T7,
+        T8,
+        T9,
+        T10,
+        T11,
+        T12,
+        T13,
+        T14,
+        T15,
+        T16,
+        T17,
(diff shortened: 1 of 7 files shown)


govert:
That looks good - can you perhaps reduce the churn on line endings and other non-functional edits?

mozcanaurelius:
@govert Thank you for your fast reply, I removed the line-ending churn and other unnecessary edits.

Sergey-Vlasov:
We do have ExtendedFunc in https://github.com/Excel-DNA/ExcelDna/blob/master/Source/ExcelDna.Integration/Registration/ExtendedFunc.cs

I was concerned that making it public can significantly increase compiled AOT add-in file size. Just tested it and the size increased only from 5.01 MB to 5.37 MB.

melihozcan1:
Okay, then I will use the already provided ExtendedFunc by making it public if that's okay @Sergey-Vlasov

Sergey-Vlasov:
> Okay, then I will use the already provided ExtendedFunc by making it public if that's okay @Sergey-Vlasov

I can try to implement it myself this week. (Plus I think ExtendedFunc should be moved to the general Integration namespace and add unit tests.)

mozcanaurelius:
Okay, I'm waiting for your PR, closing this one. Thanks. @Sergey-Vlasov

PR #844, opened 2026-04-12
Added support for AOT functions with args > 16

Sergey-Vlasov:


mozcanaurelius:
Hi, I've reviewed the PR and it solves all the problems I've encountered while working with sync. / async. functions w. more than 16 arguments for NAOT compilation. Can you review and merge this PR when you have the time? @govert  
Thank you in advance.

Source/ExcelDna.SourceGenerator.NativeAOT/Generator.cs
```diff
@@ -93,18 +93,32 @@ public static short Initialize(void* xlAddInExportInfoAddress, void* hModuleXll,
...
                 {
                     functions += $"{regHost}.MethodsForRegistration.Add({GetMethod(i)});\r\n";
-                    functions += $"typeRefs.Add(typeof({Util.MethodType(i)}));\r\n";
-                    functions += $"typeRefs.Add(typeof({Util.MethodExpressionType(i)}));\r\n";
+                    if (i.Parameters.Length <= 16)
+                    {
+                        functions += $"typeRefs.Add(typeof({Util.MethodType(i)}));\r\n";
+                        functions += $"typeRefs.Add(typeof({Util.MethodExpression(Util.MethodType(i))}));\r\n";
+                    }
+                    {
```
govert: do we need an else here ?

Source/ExcelDna.SourceGenerator.NativeAOT/Generator.cs
```diff
@@ -93,18 +93,32 @@ public static short Initialize(void* xlAddInExportInfoAddress, void* hModuleXll,
...
                     {
-                        string convertedMethodType = Util.MethodPostParameterConversionType(i);
-                        functions += $"typeRefs.Add(typeof({convertedMethodType}));\r\n";
-                        functions += $"typeRefs.Add(typeof(System.Linq.Expressions.Expression<{convertedMethodType}>));\r\n";
+                        if (i.Parameters.Length <= 16)
+                        {
+                            functions += $"typeRefs.Add(typeof({Util.MethodPostParameterConversionType(i)}));\r\n";
+                            functions += $"typeRefs.Add(typeof({Util.MethodExpression(Util.MethodPostParameterConversionType(i))}));\r\n";
+                        }
+                        {
```
govert: else here?

Sergey-Vlasov:
Our ExtendedFunc supports any number of parameters (up to 256), so we don't need a condition around its usage.

The Integration project uses both Func and ExtendedFunc, thus we pregenerate both types.

PR #849, opened 2026-05-03
Added support for loading referenced .dlls to the app domain in .NET Core

### LINKED ISSUE #795: WPF control library's resource assembly not found at runtime despite being present in output directory (worked before system reinstall)
### Description

**Environment:**
- ExcelDNA Version: [latest version, no .dna file]
- .NET Framework: [your version]
- Excel Version: [your Excel version]
- Visual Studio: [your VS version]
- Windows Version: [your Windows version]

**Problem:**
I'm developing an ExcelDNA add-in that references a WPF control library via NuGet. This control library has a dependency on a style/theme resource assembly (`Cyclone.Wpf.Themes`). 

**Critical Information: This add-in was working perfectly before I reinstalled Windows. After the system reinstall, with the same code and project configuration, the add-in now throws a FileNotFoundException for the resource assembly.**

**Important: The `Cyclone.Wpf.Themes.dll` file IS present in the bin\Debug (or bin\Release) directory after building, sitting right next to the main add-in DLL and the .xll file.** However, when the ExcelDNA add-in launches a WPF window at runtime, it throws a FileNotFoundException for this assembly.

**Error Message:**
```
System.Windows.Markup.XamlParseException
  HResult=0x80131501
  Message="设置属性"System.Windows.ResourceDictionary.Source"时引发了异常。"，行号为"3"，行位置为"10"。
  Source=PresentationFramework
  StackTrace:
   在 System.Windows.Markup.XamlReader.RewrapException(Exception e, IXamlLineInfo lineInfo, Uri baseUri)
   在 System.Windows.Markup.WpfXamlLoader.Load(XamlReader xamlReader, IXamlObjectWriterFactory writerFactory, Boolean skipJournaledProperties, Object rootObject, XamlObjectWriterSettings settings, Uri baseUri)
   在 System.Windows.Markup.WpfXamlLoader.LoadBaml(XamlReader xamlReader, Boolean skipJournaledProperties, Object rootObject, XamlAccessLevel accessLevel, Uri baseUri)
   在 System.Windows.Markup.XamlReader.LoadBaml(Stream stream, ParserContext parserContext, Object parent, Boolean closeStream)
   在 System.Windows.Application.LoadComponent(Object component, Uri resourceLocator)
   在 Cyclone.ExcelToolkit.Views.OrderDataFillView.InitializeComponent() 在 E:\Code\Cyclone\ExcelToolkit\Cyclone.ExcelToolkit\Views\Production\OrderDataFillView.xaml 中: 第 1 行

内部异常 1:
FileNotFoundException: Could not load file or assembly 'Cyclone.Wpf.Themes, Culture=neutral, PublicKeyToken=null'. 系统找不到指定的文件。
```

**Project Structure:**
```
ExcelDNA Add-in Project
├── References WPF Control Library (NuGet)
    └── WPF Control Library references Cyclone.Wpf.Themes (style/resource assembly)

Output Directory (bin\Debug):
├── MyAddIn.dll
├── MyAddIn.xll
├── MyAddIn-packed.xll
├── WpfControlLibrary.dll
├── Cyclone.Wpf.Themes.dll  ← THIS FILE EXISTS HERE
└── Other dependencies...
```

**Steps to Reproduce:**
1. Create an ExcelDNA add-in project using the latest NuGet package
2. Add a reference to a WPF control library that depends on a separate resource assembly
3. Build the project - verify that all DLLs including `Cyclone.Wpf.Themes.dll` are in the output directory
4. Create a WPF window that uses controls from the library
5. Launch the window from the Excel add-in
6. The FileNotFoundException occurs when the window tries to load resources, even though the DLL is physically present in the same directory

**What Changed:**
- **Before system reinstall**: The add-in worked perfectly with the exact same code and project structure
- **After system reinstall**: The FileNotFoundException occurs
- **No code changes** were made between the working and non-working states
- All Visual Studio, .NET Framework, and Excel versions were reinstalled to the same versions as before

**Additional Information:**
- **The Cyclone.Wpf.Themes.dll IS present in the bin\Debug directory** - this is not a build/copy issue
- The same WPF window works correctly in a standalone WPF application with the same DLL structure
- The issue only occurs when running as an ExcelDNA add-in within Excel process
- The resource dictionary is referenced in XAML like: `<ResourceDictionary Source="/Cyclone.Wpf.Themes;component/Themes/Generic.xaml"/>`
- Using Process Monitor shows Excel is not even attempting to load the DLL from the add-in directory

**Expected Behavior:**
Since all required DLLs are present in the output directory, and the add-in worked before the system reinstall, the WPF window should be able to load its resource assemblies when launched from the ExcelDNA add-in.

**Core Questions:**
1. Why can't WPF/XAML find the `Cyclone.Wpf.Themes.dll` assembly at runtime when running inside Excel via ExcelDNA, even though the file exists in the same directory as the add-in?
2. **What system-level configuration or dependency could have changed after a Windows reinstall that would affect assembly loading in ExcelDNA but not in standalone applications?**
3. Is there a Windows setting, GAC registration, or .NET configuration that ExcelDNA relies on that might not be properly configured after a fresh Windows installation?
4. Could this be related to security policies or trust levels that affect how Excel loads assemblies?

Sergey-Vlasov:


govert:
How does the LoadFrom(path) interact with the AssemblyLoadContext here? It sounded like the EnterContextualReflection call was better - that ensures the chain of loaded assemblies resolve inside the ALC. Otherwise, they might load into the default ALX, which we don't want. It might be helpful to check the ALCs for the assemblies afterwards to confirm.
If the EnterContextualReflection is the fix, then I'd rather have guidance somewhere saying when it is recommended or needed. Automatically loading failing assemblies by our own probe and LoadFrom seems wrong.

Sergey-Vlasov:
Yes, the implemented AppDomain.CurrentDomain.AssemblyResolve solution loads into the default global ALC.

EnterContextualReflection is a solution, if the add-in developer can add the code to invoke it.

Actually, we can use both. The implemented AssemblyResolve makes sure the add-in doesn’t crash by default, and if the add-in requires private ALC then adding a call to EnterContextualReflection per guidance will make it so.

PR #851, opened 2026-05-17
Added AppContext.BaseDirectory initialization for .NET Core

### LINKED ISSUE #757: AppContext.BaseDirectory returns empty string in .Net 8 or 9
the ExcelDna version is 1.8 or 1.9, the project uses AppContext.BaseDirectory to get the application directory, in the net framework , it works fine . But in Net 8 or 9 , it alwarys returns empty string. In fact, not only AppContext.BaseDirectory , but also AppDomain.CurrentDomain.BaseDirectory, Assembly.GetExecutingAssembly().Location are the same.

Image

Sergey-Vlasov:


govert:
I'm a bit nervous of whether this is something we should be setting or overriding with the xll directory. It looks like a runtime wide setting, so multiple add-ins will not be able to depend on it.

Sergey-Vlasov:
Yes, in .NET Core it is unfortunately a global runtime wide setting, so multiple add-ins will not be able to depend on it.

The proposed implementation doesn’t override anything, without it `AppContext.BaseDirectory` is just not set.

As a workaround, `Path.GetDirectoryName(ExcelDnaUtil.XllPath)` can be used instead of `AppContext.BaseDirectory`.

govert:
I'm closing this one - I don't think we should add this as a standard initialization.

